PR #36, opened 2020-08-24
Update code example for truncate component.

fabiankaegy:
The current example is too short to show the truncate effect on larger screens. Therefore I used the text from the second example in the first aswell to make sure it is long enough to always show up.

Screen Shot 2020-08-24 at 22 51 19


ItsJonQ:
@fabiankaegy Awesome! Thank you! I'll take care of that merge conflict :)

PR #45, opened 2020-09-03
Styles: Add StyleFrameProvider for iFrame rendering

ItsJonQ:
Screen Shot 2020-09-03 at 4 19 54 PM

This update adds a new component to the `@wp-g2/styles` package, `<StyleFrameProvider />`.

`<StyleFrameProvider />` is to handle component rendering in iFrames. It does this by identifying the **correct** `document.head` DOM node to synchronize styles rendered by the custom Emotion instance.

In other words... components should render perfectly whether they live in an iFrame (any iFrame) or not, **without** having to do anything (other than using `<StyleFrameProvider />`.

Example:

```jsx
<Frame>
  <StyleFrameProvider>
    <Button>Let It Go</Button>
  </StyleFrameProvider>
</Frame>
```

A special thanks to @aristath for popping in the Zoom call and pairing (🍐)! It was really helpful (and fun)

**Demo Preview:**
https://g2-git-style-frame-provider.itsjonq.vercel.app/iframe.html?id=examples-wip-contextrendering--default&viewMode=story


ItsJonQ:
Interesting!... it doesn't seem to work on production builds 🤔 .
It works for dev builds though.

Will need to continue poking!

ItsJonQ:
Fixed production (`isSpeedy`) rendering!

PR #50, opened 2020-09-08
Context: Enhance CSS targeting flow

ItsJonQ:
This update improves the `<ComponentsProvider />` to gracefully target connected components via a (context) `css` value. This special `css` value gets **compiled** and **merged** with the connected component (accepting incoming `css` props).

For example:

```jsx
const Example = () => {
  const value = {
    Card: {
      css: {
        background: 'blue',
      },
    },
    CardBody: {
      css: [
        `
          padding: 20px;
          *:hover > & {
              padding: 40px;
          }
      	`,
        ui.animation.default,
      ],
    },
  }

  return (
    <>
      <ComponentsProvider value={value}>
        <Card>
          <CardBody css={[{ border: '3px solid green' }]}>
            <Text optimizeReadabilityFor="blue">Card</Text>
          </CardBody>
        </Card>
      </ComponentsProvider>
    </>
  )
}
```

The `CardBody` will render the CSS defined in the `ComponentsProvider` as well as the locally defined `css` prop.


packages/components/src/Spacer/Spacer.js
```diff
@@ -1,5 +1,5 @@
 import { connect } from '@wp-g2/context';
-import { css, cx, toPx } from '@wp-g2/styles';
+import { css, cx, space } from '@wp-g2/styles';
```
ItsJonQ, line 2: Fix to use the `space` mixin, rather than directly multiplying the value by `4`

packages/styles/src/css.js
```diff
@@ -8,32 +8,28 @@ const { breakpoints, css: compile } = compiler;
...
-	paddingLeft: 'space',
-	paddingX: 'space',
-	paddingY: 'space',
-	paddingBlock: 'space',
-	paddingBlockEnd: 'space',
-	paddingBlockStart: 'space',
-	paddingInline: 'space',
-	paddingInlineEnd: 'space',
-	paddingInlineStart: 'space',
+	// margin: 'space',
```
ItsJonQ, line 11: Disabling for margin/padding (for now).

PR #58, opened 2020-09-24
Attempt to improve design tools performance

diegohaz:
The code is a mess. And there may be some unnecessary optimizations. But the result is great! 😆


ItsJonQ:
@diegohaz In other words... we'll built our own Redux! With context and pubsubs! 😂

ItsJonQ:
@diegohaz Thank you so much for investigating this! I'll poke around some more tomorrow!! <3

ItsJonQ (approved):
❤️ Really fascinating! Thanks @diegohaz !!!!

PR #64, opened 2020-10-13
Build: Composite TypeScript project with type generation

sirreal:
## Description

Generate TypeScript declaration files (`module.d.ts`) instead of manually maintaining them. Files remain JavaScript, TypeScript uses JSDoc and the JavaScript language.

Remove the existing manually maintained type declarations. They should not be necessary and the generated ones will be more accurate and automatically up to date.

The generated types will be put in the same place, so the published package will see no difference.

This is a _composite_ TypeScript project. The structure reflects the monorepo structure. There is a main `tsconfig.json` file at the root the points to all of the packages as references.

All of the packages extend a common `tsconfig.base.json` configuration at the project root. They have package-specific settings to locate their source and output. They reference each other based on the package dependencies.

This is very similar to the way other packages have been setup, e.g. [Gutenberg](https://github.com/WordPress/gutenberg) or [Tannin](https://github.com/aduth/tannin).

## Status

Setup is complete for the most part. The build currently fails and requires a little bit of tweaking.

I didn't integrate anything with existing setup and tooling. This leverages `tsc` directly, `tsc --build` should be run before packaging/publishing the packages.

## How to use this?

Start up the TypeScript build via `tsc` (TypeScript compiler) as follows:

```sh
node_modules/.bin/tsc --build
```

Add `--watch` to… watch 🙂 

```sh
node_modules/.bin/tsc --build --watch
```

Clean TypeScript artifacts (shouldn't be necessary):

```sh
node_modules/.bin/tsc --build --clean
```


ItsJonQ:
@sirreal Wooo! TypeScript 😍 .

Forgive my n00bness... but I'm not sure what this update does 😊 

I noticed that the previous types have been removed.
With this new workflow, would there be a way to keep some of the type descriptions and examples, like this?

https://github.com/ItsJonQ/g2/pull/64/files#diff-484fd4ffa17d9aa342211d392f5bde5d08e878c158d6da43fc31909e36d480a1L13

Thanks!

sirreal:
> I'm not sure what this update does 😊

That was my fault, I've added a description to clarify what's going on here 👍 

> I noticed that the previous types have been removed.

The goal is to generate types from source code instead of writing the code on one side and writing type declarations independently. In the case of JavaScript, this will be based on JSDoc.

There are a few issues that TypeScript has already highlighted to fix. For example, the `is.number` function is used but [does not exist on the type](https://github.com/ItsJonQ/is/blob/master/types/index.d.ts). This is a perfect example of the benefits of generating the types rather than maintaining them independent of the source 🙂

sirreal:
This is mostly just sketching so I'll close. It would take a bit of work to get this compiling and I'm unable to invest that right now.

PR #72, opened 2020-10-30
Add BlankSlate component

saramarcondes:


packages/context/src/useContextSystem.js
```diff
@@ -114,7 +114,8 @@ export function useContextSystem(props, namespace) {
 function getStyledClassName(displayName) {
 	if (!displayName || !is.string(displayName)) return '';
 
-	return `wp-components-${kebabCase(displayName)}`;
+	const kebab = kebabCase(displayName);
+	return `components-${kebab} wp-components-${kebab}`;
```
ItsJonQ, line 118: 👍 with this change. We'll need to update a bunch of snapshots to make this happy 😂 .

packages/components/src/BlankSlate/BlankSlate.js
```diff
@@ -0,0 +1,57 @@
...
+
+	const classes = cx(className, modifierClassNames);
+
+	return (
+		<Card className={classes} {...otherProps} ref={forwardedRef}>
+			{resizeListener}
+			<CardBody className={cardBodyStyles}>
+				<VStack>
+					<Heading size={1}>
+						<Icon className={iconStyles} icon={icon} inline />
```
ItsJonQ, line 46: @saramarcondes I think it may be a good idea to conditionally render the `Icon` and `Text` (description), just in case those props are not passed in. That way, we won't have empty `Icon` and `Text` nodes <3

packages/components/src/BlankSlate/BlankSlate.js
```diff
@@ -0,0 +1,57 @@
...
+		className,
+		description,
+		icon,
+		title,
+		...otherProps
+	} = useContextSystem(props, 'BlankSlate');
+
+	const [resizeListener, { width }] = useResizeAware();
+
+	console.log(width);
```
ItsJonQ, line 25: I think the width is working. Time to remove this stray `width` console.log 😊

packages/components/src/BlankSlate/BlankSlate.js
```diff
@@ -0,0 +1,57 @@
...
+import { cx } from '@wp-g2/styles';
+import { useResizeAware } from '@wp-g2/utils';
+import React from 'react';
+
+import { Card, CardBody } from '../Card';
+import { Heading } from '../Heading';
+import { Icon } from '../Icon';
+import { Text } from '../Text';
+import { VStack } from '../VStack';
+import { cardBodyStyles, iconStyles } from './BlankSlate.styles';
```
ItsJonQ, line 11: How I've been handling the imports of styles is to use:

```jsx
import * as styles from '...';
```

My thinking for that is that it's the same workflow as CSS Modules, which helps with familiarity.
This means that in the `BlankState.styles` file, the exports can just be `CardBody` or `Icon`.

And it can be used like `styles.CardBody` and `styles.Icon` within this file (just like CSS modules) :)

ItsJonQ (approved):
🚀 from me! Thanks @saramarcondes !!!!!

PR #73, opened 2020-11-02
Add dark and high contrast theme support to ThemeProvider

saramarcondes:
### Testing instructions

Checkout the branch locally and pass `darkTheme`, `highContrastTheme` and `darkHighContrastTheme` to the `ThemeProvider` in `.storybook/preview.js`. Pass some different variables and ensure they're being applied correctly with the expected specificity.

### DIFF
--- a/.storybook/themes/blueberry.js
+++ b/.storybook/themes/blueberry.js
@@ -1,6 +1,6 @@
 import { createTheme } from '@wp-g2/styles';
 
-const theme = createTheme(({ get, space, ...theme }) => {
+const theme = createTheme(({ get, space, theme }) => {
 	const COLOR_PROPS = {
 		colorAdmin: '#3858E9',
 		colorText: '#3a3b3c',
--- a/packages/create-styles/src/components/ThemeProvider/ThemeProvider.js
+++ b/packages/create-styles/src/components/ThemeProvider/ThemeProvider.js
@@ -1,4 +1,4 @@
-import { repeat } from '@wp-g2/utils';
+import { mergeRefs, repeat } from '@wp-g2/utils';
 import React, { useRef } from 'react';
 
 import {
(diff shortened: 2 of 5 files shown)


packages/create-styles/src/components/ThemeProvider/ThemeProvider.js
```diff
@@ -49,19 +56,48 @@ function ThemeProvider({
...
+	darkHighContrastTheme = {},
 	...props
 }) {
 	/**
 	 * Hydrates global styles (via injectGlobal). This is necessary as there may
 	 * be a chance that <ThemeProvider /> renders before any other (styled)
 	 * component. Injecting global styles early in this manner allows for
 	 * the initial render of theme styles (which also uses injectGlobal)
-	 * to be sequences correctly.
+	 * to be sequenced correctly.
```
ItsJonQ, line 69: Thank you 🙏

ItsJonQ (approved):
@saramarcondes 🚀 from me! Tested it in Storybook, and the new theme modes are working as expected, including local settings (`isGlobal={false}`) and global settings.

(Not a blocker) The thing I'm not sure of is it looks like the `themeStyles` object is newly created every render. Previously, it was the instance provided by the substate store (powered by `zustand`).

It doesn't appear to be a problem. But something to keep an eye on.

Lemme know if you have any thoughts. If you think it's 👍 , then I'll merge this up!

saramarcondes:
@ItsJonQ 

> The thing I'm not sure of is it looks like the `themeStyles` object is newly created every render

Is this a new thing from my changes? I'm not really sure I would know where to start to try to address this... do you have any pointers?

Also... have you observed a performance deficit due to this?

ItsJonQ:
> Is this a new thing from my changes?

@saramarcondes I believe so :)

>  I'm not really sure I would know where to start to try to address this...

The theme object was created/managed with this:

https://github.com/ItsJonQ/g2/blob/a5f1fb1a74f78c72b0678e3e5ce919c367968960/packages/create-styles/src/components/ThemeProvider/ThemeProvider.utils.js#L141

That `theme` object is only changed when the values are updated. Otherwise, it should be the same object being returned.

With this update, we're getting the 4 `theme` objects and merging them on every render.

> Also... have you observed a performance deficit due to this?

Yes and no 😂 .

Yes, if you needed to update/re-render the `ThemeProvider` often (e.g. changing the `colorAdmin` variable using a color picker).

However! There's an interesting thing about G2's ThemeProvider. As you've discovered, because of how G2 handles theming/variables, it doesn't require to you to put content (`children`) inside of `ThemeProvider`. Which means, there's no gigantic React node-tree walk down/up per render.

This only applies to "global" updates, not scoped theme updates (controlled by the `isGlobal` prop).

☝️ With this being said... we may be okay for now :).

Lemme know what you think. If this makes sense, then we can merge it up

.storybook/themes/blueberry.js
```diff
@@ -1,6 +1,6 @@
 import { createTheme } from '@wp-g2/styles';
 
-const theme = createTheme(({ get, space, ...theme }) => {
```
ItsJonQ, line 3: Thank you for fixing this 🙈

Issue #29, opened 2020-08-19
Modal: Show/hide mechanic needs to be easier to use


ItsJonQ:
At the moment, there isn't a way to toggle the `visible` state of a `<Modal />`. You can set the initial `visible`, but it cannot be toggled after the fact.

Right now, Modals are bound to a singular Trigger target. This is typically okay if you have an action that is associated with a Modal, like

```
Show More Settings -> More settings modal
```

This is less ideal when there are multiple items that require (essentially) the same modal, for example:

```
[ Photo, Photo, Photo, Photo ]
```

Clicking any of the photos should bring up a Modal.

Right now, each Photo requires their own Modal. Instead, it should be more like:

```
[ Photo, Photo, Photo, Photo ]
[ PhotoModal ]
```

PR #77, opened 2020-11-04
Add controlled modal example

saramarcondes:
### Testing instructions

Run storybook and visit the controlled modal example and ensure both buttons cause the modal to open and that focus is returned to them. Likewise, open the default modal example and ensure it still behaves as expected.

Closes https://github.com/ItsJonQ/g2/issues/29


ItsJonQ (approved):
🚀 from me! Thank you @saramarcondes . This implementation of controlled state works great ❤️

Issue #76, opened 2020-11-04
UnitControl: Allow for values like `calc()`


ItsJonQ:
In addition to units, and valid CSS values (like `auto`), it should also be able to handle valid CSS values like `calc()`.

PR #80, opened 2020-11-04
UnitInput: Improve cssProp validation to allow for values like calc()

ItsJonQ:
Screen Shot 2020-11-04 at 4 34 54 PM

This update improves the CSS prop validation util function that's used in `UnitInput`. This fix allows for values like `calc()`, `var()`, and other native CSS functions.

### 🕹 Testing

* Run `yarn start` to fire up Storybook
* Go to the [UnitInput story](http://localhost:6006/iframe.html?id=components-unitinput--default)
* Add whatever (valid) `calc()` value to one of the inputs. Example `calc(1px + 2px)`

Resolves: https://github.com/ItsJonQ/g2/issues/76

### DIFF
--- a/packages/utils/src/unitValues.js
+++ b/packages/utils/src/unitValues.js
@@ -24,13 +24,15 @@ export const isValidCSSValueForProp = (prop, value) => {
 
 	if (is.undefined(computedStyleMap[prop])) return true;
 
-	// Reset
+	// 1. Reset current style value.
 	computedStyleMap[prop] = '';
-
+	// 2. Cache current style value for validation (may not be an empty string).
+	const current = computedStyleMap[prop];
+	// 3. Apply next value.
 	const next = getCSSValue(value);
 	computedStyleMap[prop] = next;
-
-	return computedStyleMap[prop].includes(value);
+	// 4. Check to see if next value was correctly applied.
+	return current !== computedStyleMap[prop];
 };
 
 export const isValidNumericUnitValue = (value) => {
--- a/packages/components/src/UnitInput/index.d.ts
+++ b/packages/components/src/UnitInput/index.d.ts
@@ -1,0 +1,1 @@
+export * from '../../types/components/UnitInput';


ItsJonQ:
## Test URL

https://g2-git-update-unit-input-support-calc.itsjonq.vercel.app/iframe.html?id=components-unitinput--default&viewMode=story

packages/utils/src/unitValues.js
```diff
@@ -24,13 +24,15 @@ export const isValidCSSValueForProp = (prop, value) => {
 
 	if (is.undefined(computedStyleMap[prop])) return true;
 
-	// Reset
+	// 1. Reset current style value.
 	computedStyleMap[prop] = '';
-
+	// 2. Cache current style value for validation (may not be an empty string).
+	const current = computedStyleMap[prop];
```
saramarcondes, line 30: I'm confused what this is doing. To me this reads like this:

```
let computedStyleMap = { 'a': ... }
let prop = 'a'

// this overwrites 'a' to be an empty string
computedStyleMap[prop] = '';

// now current will always be an empty string
const current = computedStyleMap[prop];
```

Did we do this in the wrong order? Do we need to cache the "current" before resetting (and why reset at all if we always overwrite it in step three?)

ItsJonQ: > Do we need to cache the "current" before resetting (and why reset at all if we always overwrite it in step three?)

We do! The CSSOM is interesting 🤣 .

Check it out.

Open up Inspector/Dev tools/console in your browser.

Run this:

```js
document.body.style.border = 'none'
```

Cool. We're explicitly disabling the border for the `body`.

Let's check what the computed value is:

```js
window.getComputedStyle(document.body).border
```

Is the value `none`?

**Nope**!

It's

```
0px none rgb(36, 41, 46)
```

🙃 .

So wait... What is the default value of `document.body.style.border` then?

Let's check...

Run this:

```js
document.body.style.border = '1px solid red'
// Result
window.getComputedStyle(document.body).border
// "1px solid rgb(255, 0, 0)"
```

So far so good. Let's unset it (or remove it).

```js
document.body.style.border = ''
```

The computed style should now be empty or `null`... right?

**Nope!**

```js
window.getComputedStyle(document.body).border
// "0px none rgb(36, 41, 46)"
```

🙃 

---

The value to assign to the style key may not equal to the final computed value (CSSOM reasons).
Therefore... What I tried to do was:

1. Reset
2. Cache the default reset value.
3. Apply the "next" value
4. See if it's different vs. the default value.

Hopefully this makes sense 🙏

saramarcondes: gotcha! That makes a lot of sense, thank you!

saramarcondes (approved):
LGTM

Issue #82, opened 2020-11-05
Style System: Enable prop interpolation in styled component


ItsJonQ:
At the moment, the Style system doesn't allow for [prop interpolation](https://emotion.sh/docs/styled#changing-based-on-props) for either `css` or `styled`.

This is because G2's `styled` is a simple wrapper for the base `css` function, which only accepts `object`, `string`, or an Array of the `object`|`string` as arguments. This is (mostly) the default behaviour of Emotion's core [css function](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L78).

However, with `styled` being an HOC wrapper, we may be able to do enhance how CSS is compiled.

The benefit of this is that it provides the prop interpolation feature that most libraries support with their styled components - making it easier for others to adopt, understand, and migrate to.

PR #84, opened 2020-11-05
Style System: Enable prop interpolation for styled components

ItsJonQ:
This update allows for styled components (created with `styled.div'...'`) to support [prop interpolation](https://emotion.sh/docs/styled#changing-based-on-props).

This allows for style workflows similar to that of Emotion and Styled Components.

### ✨ Example

In the example below, we're able to adjust the styles of the `FriendlyExample` component through mixins and inline functions.
These functions output a different style rule based on `props` that the component receives.

```jsx
const happyMixin = (props) => {
	return props.happy ? `color: green` : `color: blue`;
};

const niceMixin = (props) => {
	return props.nice ? `background-color: yellow` : `background-color: #ddd`;
};

const FriendlyExample = styled.div`
	color: #aaa;
	${happyMixin};
	${niceMixin};
	font-size: 40px;
	font-weight: bold;
	padding: 20px;
	${(props) => props.title && `font-size: 90px`}
`;

const App = () => {
	return (
		<FriendlyExample happy={false} nice={false} title="Weee">
			Hello
		</FriendlyExample>
	);
};
```

### 🤔 How does this work?

This is achieved by resolving the "tagged template literal" arguments - a combination of `string` and `function` objects that the `styled` component receives (via the backtick).

This strategy is the same as the one from libraries like Styled Components:
https://mxstbr.blog/2016/11/styled-components-magic-explained/

This update handles this compiling mechanic with this dedicated function:

```js
export function compileInterpolatedStyles(interpolatedStyles, props) {
	const resolvedInterpolatedStyles = interpolatedStyles.map((a) =>
		is.function(a) ? a(props) : a,
	);

	const [cssChunk, ...interpolatedCSSRule] = resolvedInterpolatedStyles;

	const rules = [...cssChunk].slice(0, -1);
	const end = [...cssChunk].slice(-1);

	const compiledStyles = rules
		.map((rule, index) => rule + interpolatedCSSRule[index])
		.join('')
		.concat(end);

	return compiledStyles;
}
```

---

### 🕹 Testing

* Play around with/test out the [`styledComponentInterpolation.stories.js` story](http://localhost:6006/iframe.html?id=create-styles-styledcomponent-stringinterpolation--default&viewMode=story)
* Alternatively, dive into an existing story and create a styled component with mixins.
* Ensure it works
* Ensure existing component styles aren't broken (🤣)

---

Resolves: https://github.com/ItsJonQ/g2/issues/82

### DIFF
--- a/packages/create-styles/src/createStyleSystem/utils.js
+++ b/packages/create-styles/src/createStyleSystem/utils.js
@@ -127,3 +127,32 @@ export function getDisplayName(Component) {
 
 	return displayName;
 }
+
+/**
+ * Resolves and compiles interpolated CSS styles for styled-components.
+ * Allows for prop (function) interpolation within the style rules.
+ *
+ * For more information on tagged template literals, check out:
+ * https://mxstbr.blog/2016/11/styled-components-magic-explained/
+ *
+ * @param {Array<string,function>} interpolatedStyles The interpolated styles from a Styled component.
+ * @param {Object} props Incoming component props.
+ * @returns {string} Compiled CSS style rules.
+ */
+export function compileInterpolatedStyles(interpolatedStyles, props) {
+	const resolvedInterpolatedStyles = interpolatedStyles.map((a) =>
+		is.function(a) ? a(props) : a,
+	);
+
+	const [cssChunk, ...interpolatedCSSRule] = resolvedInterpolatedStyles;
+
+	const rules = [...cssChunk].slice(0, -1);
+	const end = [...cssChunk].slice(-1);
+
+	const compiledStyles = rules
+		.map((rule, index) => rule + interpolatedCSSRule[index])
+		.join('')
+		.concat(end);
+
+	return compiledStyles;
+}
--- a/packages/create-styles/src/createStyleSystem/utils.js
+++ b/packages/create-styles/src/createStyleSystem/utils.js
@@ -140,19 +140,9 @@ export function getDisplayName(Component) {
  * @returns {string} Compiled CSS style rules.
  */
 export function compileInterpolatedStyles(interpolatedStyles, props) {
-	const resolvedInterpolatedStyles = interpolatedStyles.map((a) =>
+	const compiledStyles = interpolatedStyles.map((a) =>
 		is.function(a) ? a(props) : a,
 	);
-
-	const [cssChunk, ...interpolatedCSSRule] = resolvedInterpolatedStyles;
-
-	const rules = [...cssChunk].slice(0, -1);
-	const end = [...cssChunk].slice(-1);
-
-	const compiledStyles = rules
-		.map((rule, index) => rule + interpolatedCSSRule[index])
-		.join('')
-		.concat(end);
 
 	return compiledStyles;
 }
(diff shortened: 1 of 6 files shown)


packages/create-styles/src/createStyleSystem/utils.js
```diff
@@ -127,3 +127,32 @@ export function getDisplayName(Component) {
...
+
+	const [cssChunk, ...interpolatedCSSRule] = resolvedInterpolatedStyles;
+
+	const rules = [...cssChunk].slice(0, -1);
+	const end = [...cssChunk].slice(-1);
+
+	const compiledStyles = rules
+		.map((rule, index) => rule + interpolatedCSSRule[index])
+		.join('')
+		.concat(end);
```
saramarcondes, line 155: Can you briefly describe what's going on here? I'm having a hard time grokking it.

ItsJonQ: @saramarcondes To be honest... I'm not sure 😂 . It was code [I've written before and borrowed](https://github.com/helpscout/fancy/blob/1.1.0/src/utilities/primitives.js#L39) for this update.

For context, the code I've written before needed to do the same CSS-in-JS interpolating thing.

---

I **believe** it's something like this!

Incoming, the `interpolatedStyles` may look like this example:

```
interpolatedStyles = [['a', 'b', ';'], fn, fn]
```

`resolvedInterpolatedStyles` resolves the functions (`fn`), so it becomes this:


```
resolvedInterpolatedStyles = [['a', 'b', ';'], 'one', 'two']
```

Next, we destructure:

```
const [cssChunk, ...interpolatedCSSRule] = resolvedInterpolatedStyles;
// cssChunk = ['a', 'b', ';']
// interpolatedCSSRule = ['one', 'two']
```

Then, we split off `rules` and `end`:

```
rules = ['a', 'b']
end = [';']
```

Lastly, we combine it all together in that `.map` + `.join`, so that we get:

```
aonebtwo;
```

(The actual style rules would have `;` in between the current chunks.)

---

Hope this helps!

saramarcondes (changes requested):
I've tried debugging this issue, but styles are breaking with this change:

https://g2-git-try-styled-component-fn-interpolation.itsjonq.vercel.app/iframe.html?id=components-checkbox--default

https://g2-git-try-styled-component-fn-interpolation.itsjonq.vercel.app/iframe.html?id=components-colorcircle--default

Those are just the two examples I could find... I believe it has something to do with the `presets` not interpolating correctly. I couldn't figure out how to effectively debug this.

ItsJonQ:
@saramarcondes Oh dang!! Thank you so much for catching those. I'll have to dive back in to debug 😅 .
I'm not sure what may be causing this either, haha

ItsJonQ:
@saramarcondes Thank you again for catching those checkbox/radio components! They had the answer for why this broke.

I just pushed a solution 👍 . The logic is simplified **a lot**.

We don't need to re-order + stringify the compiled styles (what strange array logic you noticed before).
The reason is because the `css(...)` function will do that for us. We just need to pass it values that it understands.
Therefore, we **just need to resolve inline functional mixins**.

TLDR: Fire functions with props. Let `css()` handle the rest ✌️

PR #87, opened 2020-11-05
Add mad libs example for inline text input

saramarcondes:


ItsJonQ:
@saramarcondes So cool to see this work 😍 

https://g2-git-add-mad-libs.itsjonq.vercel.app/iframe.html?id=components-textinput--mad-libs&viewMode=story

Issue #86, opened 2020-11-05
Select: Support multiple selection


ItsJonQ:
We need to enable the `Select` component to support the selection of multiple values.

See the `@wordpress/components` implementation for ideas:
https://github.com/WordPress/gutenberg/blob/master/packages/components/src/select-control/index.js#L66

Issue #85, opened 2020-11-05
Select: Support grouping for values passed in `options` prop


ItsJonQ:
The `Select` component can render `<option />` components in 2 ways:

* Passing options as `children` (more flexible)
* Passing options (`Array<object>`) as `options` prop (more automated)

We should support grouping (via `optgroup`) for values passed in via the `options` prop.

At the moment, `optgroup` grouping is only achievable by manually constructing the markup via `children`.

PR #88, opened 2020-11-05
Select: Add option grouping, placeholder, and multiple support

ItsJonQ:
This update enhances the `Select` component to support grouping (when passing in values via the `options` prop), `placeholder` rendering, and multi-select via the `multiple` attribute.

In addition to adding these features, `Select` was featured to use the hook-based rendering approach (via `useSelect`). Type definitions have also been updated.

### ✋ Placeholder

Screen Shot 2020-11-05 at 3 22 27 PM

A "placeholder" UI can be rendered by providing a `placeholder` (`string`) value.

### 🍇 Groups

Screen Shot 2020-11-05 at 3 22 30 PM

Automatic `optgroup` rendering can be done by providing `Select` with `options` that look like this:

```js
	const options = [
		{
			label: 'Frozen',
			options: [
				{
					label: 'Frozen Heart',
					value: 'frozen-heart',
				},
				{
					label: 'Do You Want To Build A Snowman?',
					value: 'do-you-want-to-build-a-snowman',
				},
				{
					label: 'First Time In Forever',
					value: 'first-time-in-forever',
				},
			],
		},
		{
			label: 'Frozen 2',
			options: [
				{
					label: 'All Is Found',
					value: 'all-is-found',
				},
				{
					label: 'Some Things Never Change',
					value: 'some-things-never-change',
				},
				{
					label: 'Into The Unknown',
					value: 'into-the-unknown',
				},
			],
		},
	];
```

### 🔢 Multi-Select

Screen Shot 2020-11-05 at 4 01 37 PM

Multi-select can be enabled by passing `Select` a `multiple` prop. This uses the native HTML `select` multiple feature.

---

### 🕹 Testing

Check out the `Select / Groups` and `Select / Multiple` stories.

https://g2-git-update-select-option-groups-and-multiple.itsjonq.vercel.app/?path=/story/components-select--groups

---

Resolves https://github.com/ItsJonQ/g2/issues/86
Resolves https://github.com/ItsJonQ/g2/issues/85

### DIFF
--- a/packages/components/src/Select/Select.utils.js
+++ b/packages/components/src/Select/Select.utils.js
@@ -1,0 +1,35 @@
+import { is } from '@wp-g2/utils';
+import React from 'react';
+
+const isGroup = (option) => {
+	return is.object(option) && is.array(option.options);
+};
+
+const renderOption = (option = {}, index) => {
+	const { disabled, id, label, value } = option;
+	const key = id || value || index;
+
+	return (
+		<option disabled={disabled} key={key}>
+			{label}
+		</option>
+	);
+};
+
+export const renderContent = ({ children, options }) => {
+	if (children) return children;
+
+	return options.map((option, index) => {
+		if (isGroup(option)) {
+			const groupKey = option.id || option.label || index;
+
+			return (
+				<optgroup key={groupKey} label={option.label}>
+					{option.options.map(renderOption)}
+				</optgroup>
+			);
+		}
+
+		return renderOption(option, index);
+	});
+};
--- a/packages/components/src/Select/Select.utils.js
+++ b/packages/components/src/Select/Select.utils.js
@@ -10,7 +10,7 @@ const renderOption = (option = {}, index) => {
 	const key = id || value || index;
 
 	return (
-		<option disabled={disabled} key={key}>
+		<option disabled={disabled} key={key} value={value}>
 			{label}
 		</option>
 	);
--- a/packages/components/types/components/Select.d.ts
+++ b/packages/components/types/components/Select.d.ts
@@ -8,9 +8,9 @@ import { BaseFieldProps } from './BaseField';
 
 type SelectOption = {
 	id?: string;
-	value?: string;
+	value: string;
 	disabled?: boolean;
-	label?: string;
+	label: string;
 };
 
 type SelectOptionGroup = {
@@ -35,6 +35,10 @@ export declare type SelectProps = Omit<
 		 * ```
 		 */
 		isSubtle?: boolean;
+		/**
+		 * Enables selection of multiple values.
+		 */
+		multiple?: boolean;
 		/**
 		 * Options to render within `Select`.
 		 *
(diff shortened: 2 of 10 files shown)


packages/components/types/components/Select.d.ts
```diff
@@ -6,6 +6,19 @@ import {
...
+	id?: string;
+	value?: string;
+	disabled?: boolean;
+	label?: string;
+};
+
+type SelectOptionGroup = {
+	id?: string;
+	label?: string;
+	options?: Array<SelectOption>;
```
saramarcondes, line 19: I understand why `SelectOption`'s properties are optional but I think `options` and `label` at least are required, no? Personally I would make `id` required too (for both) even though that slightly augments the requirements for the consumer because it absolutely prevents the consumer from programming an unsafe select.

packages/components/src/Select/Select.utils.js
```diff
@@ -0,0 +1,35 @@
...
+
+	return (
+		<option disabled={disabled} key={key} value={value}>
+			{label}
+		</option>
+	);
+};
+
+export const renderContent = ({ children, options }) => {
+	if (children) return children;
```
saramarcondes, line 20: I don't fully understand how/where `children` would come from here.

packages/components/types/components/Select.d.ts
```diff
@@ -30,16 +43,53 @@ export declare type SelectProps = Omit<
...
+		 * 				},
+		 * 			],
+		 * 		},
+		 * 	];
+		 *
+		 * 	return <Select options={options} />
+		 * }
+		 * ```
+		 */
+		options?: Array<SelectOption> | SelectOptionGroup;
```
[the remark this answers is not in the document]
ItsJonQ: Lol yes. Thank you 🙏

packages/components/src/Select/Select.utils.js
ItsJonQ: `children` would come from something like this:

```jsx
<Select>
  <option>...</option>
  <option>...</option>
  <option>...</option>
</Select>
```

☝️ a manually written collection of `options`, rather than using the `options` prop.

packages/components/types/components/Select.d.ts
ItsJonQ: > I think options and label at least are required, no

That's a good point! I'll make that update 🙏 .

> Personally I would make id required too (for both)

That was something I wasn't sure about 🤔 . The default HTML use IDs for `optgroup` and `option` in that way.

I figured I'd provide it as an option for the consumer since data typically flows into Select-like components with that structure (which would be prioritized for the React `key` prop).

packages/components/src/Select/Select.utils.js
saramarcondes: > a manually written collection of option

Gotcha! I didn't think that the component would work that way but it makes sense.

saramarcondes (approved):
🤠

PR #92, opened 2020-11-06
Proof of concept for rtlcss alternative to cssjanus

saramarcondes:
### Testing instructions

Boot up storybook and find the `html` element. Add a `dir="rtl"` attribute to it and verify that RTL support is working as expected (compare against using `cssjanus` to check that everything looks the same.

### DIFF
--- a/packages/context/src/__tests__/ContextSystemProvider.test.js
+++ b/packages/context/src/__tests__/ContextSystemProvider.test.js
@@ -44,7 +44,7 @@ describe('props', () => {
 
 		const contextValue = {
 			Olaf: {
-				css: [{ background: 'white' }, `padding: 20px`],
+				css: [{ background: 'white' }, `padding: 20px;`],
 			},
 		};
 
--- a/packages/create-styles/src/createCompiler/plugins/rtl.js
+++ b/packages/create-styles/src/createCompiler/plugins/rtl.js
@@ -4,11 +4,12 @@
  */
 
 import cssjanus from 'cssjanus';
+import rtlcss from 'rtlcss';
 
-let isRtl = false;
-if (typeof window !== 'undefined') {
-	isRtl = window?.document?.documentElement?.dir === 'rtl';
-}
+let isRtl = true;
+// if (typeof window !== 'undefined') {
+// 	isRtl = window?.document?.documentElement?.dir === 'rtl';
+// }
 
 // https://github.com/thysultan/stylis.js#plugins
 const STYLIS_CONTEXTS = {
@@ -36,7 +37,7 @@ export const STYLIS_PROPERTY_CONTEXT = STYLIS_CONTEXTS.PREPARATION;
  */
 function stylisRTLPlugin(context, content) {
 	if (context === STYLIS_PROPERTY_CONTEXT) {
-		return isRtl ? cssjanus.transform(content) : undefined;
+		return isRtl ? rtlcss.process(content) : undefined;
 	}
 }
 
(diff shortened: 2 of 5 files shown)


packages/context/src/__tests__/ContextSystemProvider.test.js
```diff
@@ -44,7 +44,7 @@ describe('props', () => {
 
 		const contextValue = {
 			Olaf: {
-				css: [{ background: 'white' }, `padding: 20px`],
+				css: [{ background: 'white' }, `padding: 20px;`],
```
saramarcondes, line 47: This is something to be aware of. Originally this test broke on a syntax error from PostCSS I believe. One small difference here between `cssjanus` and `rtlcss`.

packages/create-styles/src/createCompiler/plugins/rtl.js
```diff
@@ -4,11 +4,12 @@
...
+import rtlcss from 'rtlcss';
 
-let isRtl = false;
-if (typeof window !== 'undefined') {
-	isRtl = window?.document?.documentElement?.dir === 'rtl';
-}
+let isRtl = true;
+// if (typeof window !== 'undefined') {
+// 	isRtl = window?.document?.documentElement?.dir === 'rtl';
+// }
```
saramarcondes, line 12: Temporary to make testing easier. You still have to add `dir="rtl"` to the `html` element in storybook.

ItsJonQ:
Screen Shot 2020-11-06 at 1 03 44 PM

Holy smokes! It works 🎉 !!!

@saramarcondes I'm 💯 down with this solution. If you're okay with it, feel free to tidy this PR up and we'll get it merged 🙌

ItsJonQ:
Oh dang. I just noticed the browser version of `rtlcss.js` uses the full `postcss` library.

Which is about 20kb gzipped 😱 
https://bundlephobia.com/result?p=postcss@8.1.6

That feels really heavy 😅 

CSSJanus comes in at 1.7kb gzipped:
https://bundlephobia.com/result?p=cssjanus@2.0.1

packages/create-styles/src/createCompiler/plugins/rtl.js
ItsJonQ: Awesome :). This definitely made it easier.

Feel free to revert this when you're ready 🎉

packages/context/src/__tests__/ContextSystemProvider.test.js
ItsJonQ: I poked around, checking the compiled CSS for various components. I don't the think ending `;` affects anything.
The only things this *may* affect would be other stylis plugins (like CSS variables or specificity).

It doesn't appear to be the case 👍

ItsJonQ:
From what I gather... the only concern with this approach would be the bundle size (which is a valid concern).

I poked around, and it doesn't look like `rtlcss` or `postcss` introduces any side-effects (e.g. doing strange CSS things in the global space). Thank goodness for that.

It looks like this may be our only option for now (besides re-writing CSSJanus).
The bundle size is unfortunate, but it may be trade off we have to make.

Because this RTL plugin is a tiny module of `create-style`, the RTL parser can be something we can (very) easily swap out in the future.

With that said, **I'm okay moving forward with this approach**.

@saramarcondes What do you think :)

saramarcondes:
@ItsJonQ Sounds good to me!

ItsJonQ (approved):
🚀 from me!!! Thank you @saramarcondes 🙏

Issue #89, opened 2020-11-05
Styles: Ensure (automatic) variable fallbacks can render


ItsJonQ:
Originally discussed in the `stylis-plugin-css-variables` repo:

https://github.com/ItsJonQ/stylis-plugin-css-variables/issues/2

ItsJonQ:
I explored this a bit this morning. I think I may have an idea of how to solve this 🙏 .

It involves creating a `Map` that lives in memory where the custom variables (declared at `:root`) are stored.
That way, variables can be grabbed from this `Map` (which should work in IE) vs. `window.getComputedStyle(...).getPropertyValue('--css-variable')`, which does not.

ItsJonQ:
Ohhh boy! Exciting update for this.

I have a working solution ✨ 

**Tested on IE11** via VirtualBox:

<img width="970" alt="Screen Shot 2020-11-08 at 10 16 10 AM" src="https://user-images.githubusercontent.com/2322354/98469564-6c4ddb80-21ae-11eb-957a-61531b6ad3be.png">

(Note the fallback values on the right)

PR #94, opened 2020-11-08
Style System: Improve automated variable fallbacks

ItsJonQ:
This update fixes the CSS variable fallback handling for IE11.

This solution has been tested and confirmed working in IE11 (via VirtualBox):

Screen Shot 2020-11-08 at 10 16 10 AM

### 🙈 The Problem

The [previous plugin solution](https://github.com/ItsJonQ/stylis-plugin-css-variables) relied on an unsupported CSSOM function to retrieve the CSS variable value from `:root` ([more details here](https://github.com/ItsJonQ/stylis-plugin-css-variables/issues/2)).

For example, let's pretend this style exists:

```css
:root {
  --color: red;
}
```

The previous solution attempted to retrieve the `--color` value with this:

```js
getComputedStyle(document.documentElement).getPropertyValue('--color')
```

In IE11, this would result in an empty string (`''`) because IE11 doesn't understand what `--color` even is.

### 🏪 Alternative Root "Store"

Instead of relying on the CSS `:root {}` object, I replicated the same mechanic by establishing a "virtual" store that exists in JS. When the Style System is created, the initial CSS variables will be used to create and establish this "root store".

When the Stylis plugin goes to work, the parser will be able to retrieve the fallback values from this virtual root store (instead of using CSS `:root {}`).

### 🔌 Plugin

The majority of the plugin code was copied over from [stylis-plugin-css-variables](https://github.com/ItsJonQ/stylis-plugin-css-variables). Changes were mainly around adding the virtual store for variable fallback resolution.

### 🤝 Trade offs

This automated solution only supports variables that are:

1. Designed to live in the `:root {}` level
2. initially passed into `createStyleSystem`

That means locally scoped variables will not be resolved. Also, it (currently) won't resolve variables being defined/modified **after the fact**.

For the purposes of G2's Component system, I think it's okay if we only support values defined in our variable system. It may encourage users to work with those values, rather than define custom values of their own.

This does affect theming though. The virtual root store won't know how to re-resolve and re-render new theme values that are being pass into `<ThemeProvider />`.

Perhaps this is okay? At minimum, the UI can render for IE11 (which is great). However, it won't be fully featured compared to other browsers (e.g. theming, night mode, etc...)

---

### 🕹 Testing

(The easiest way to test)
Fire up local Storybook and inspect element. If you can see the variable fallbacks, then the system is working 👍 .

Screen Shot 2020-11-08 at 10 36 45 AM

As mentioned above, it's using the virtual "root store" strategy rather than `getComputedStyle().getPropertyValue()`, which means it will work in IE11.

Note: The CSS variable fallback handling are enabled in `develop` mode. In production, the plugin only kicks in for IE11.

---

Resolves https://github.com/ItsJonQ/g2/issues/89


packages/create-styles/src/createCompiler/plugins/cssVariables.js
```diff
@@ -0,0 +1,69 @@
...
+		length,
+		type,
+	) => {
+		// Skip generating CSS variable fallbacks for supported browsers
+		if (skipSupportedBrowsers && isNativeSupport) return;
+
+		// Borrowed guard implementation from:
+		// https://github.com/Andarist/stylis-plugin-extra-scope/blob/master/src/index.js#L15
+		/* istanbul ignore next */
+		if (context !== STYLIS_CONTEXTS.SELECTOR_BLOCK || type === 107) {
```
saramarcondes, line 52: Nit: can we give `107` a name?

packages/create-styles/src/cssCustomProperties/createRootStore.js
```diff
@@ -0,0 +1,97 @@
...
+	 * @param {string} key The key to retrieve.
+	 * @returns {string} The value.
+	 */
+	get = (key) => this.state[key];
+
+	/**
+	 * Retrieves the current state.
+	 * @returns {object} The state.
+	 */
+	getState = () => ({ ...this.state });
```
saramarcondes, line 26: Rather than spreading the state each time it's retrieved I wonder if we shouldn't just freeze it when we set it?

packages/create-styles/src/cssCustomProperties/createRootStore.js
```diff
@@ -0,0 +1,97 @@
...
+
+		return this.state;
+	};
+
+	/**
+	 * Updates the state.
+	 * @param {object} next The next state to merge into the current state.
+	 */
+	_updateState = (next = {}) => {
+		this.state = { ...this.state, ...next };
```
saramarcondes, line 45:
```suggestion
		this.state = Object.freeze({ ...this.state, ...next });
```

packages/create-styles/src/cssCustomProperties/createRootStore.js
```diff
@@ -0,0 +1,97 @@
+import { getPropValue } from './getPropValue';
+import { hasVariable, isCustomProperty } from './utils';
+
+/**
+ * Stores CSS config variables that are expected to be added to :root.
+ * This is used for CSS variable fallback handling for IE 11.
+ */
```
saramarcondes, line 7: Type correction: `state` is probably better typed as `Record<string, string>` rather than a plain `object` type. That should be corrected throughout and you can alias it using `@typedef {Record<string, string>} State` if you'd like (probably for the best).

packages/create-styles/src/cssCustomProperties/getPropValue.js
```diff
@@ -0,0 +1,61 @@
+import { sanitizeParens, VAR_REG_EXP } from './utils';
+
+/**
+ * Interprets and retrieves the CSS property and value of a declaration rule.
+ *
+ * @param {string} declaration A CSS declaration rule to parse.
+ * @param {object} rootStore A store for CSS root variables.
+ * @returns {Array<string, ?string>} [prop, value] parsed from the declaration.
```
saramarcondes, line 8: Nit: I think `?` implies `null` or `undefined` but if I'm not mistaken it's just `undefined` that is possible aside from `string` here so we can refine this a little bit.

```suggestion
 * @returns {Array<string, string | undefined>} [prop, value] parsed from the declaration.
```

packages/create-styles/src/cssCustomProperties/transformContent.js
```diff
@@ -0,0 +1,89 @@
+import { memoize } from '@wp-g2/utils';
+
+import { getPropValue } from './getPropValue';
+import { hasVariable, isCustomProperty } from './utils';
+
+/**
+ * Interprets and retrieves the CSS fallback value of a declaration rule.
+ *
+ * @param {string} declaration A CSS declaration rule to parse.
+ * @param {object} rootStore A store for CSS root variables.
```
saramarcondes, line 10: This can be typed as `RootStore` by importing it `@param {import('./createRootStore').RootStore}`

```suggestion
 * @param {import('./createRootStore').RootStore} rootStore A store for CSS root variables.
```

packages/create-styles/src/cssCustomProperties/transformContent.js
```diff
@@ -0,0 +1,89 @@
...
+
+	return value ? [prop, value].join(':') : undefined;
+}
+
+/**
+ * Parses the incoming content from stylis to add fallback CSS values for
+ * variables.
+ *
+ * @param {string} content Stylis content to parse.
+ * @param {object} rootStore A store for CSS root variables.
```
saramarcondes, line 27: Same here.

packages/create-styles/src/cssCustomProperties/createRootStore.js
```diff
@@ -0,0 +1,97 @@
...
+	 * Updates the state.
+	 * @param {object} next The next state to merge into the current state.
+	 */
+	_updateState = (next = {}) => {
+		this.state = { ...this.state, ...next };
+	};
+
+	/**
+	 * Resolves potential CSS variables that may exist within the state's values.
+	 * This function cycles a maximum of 3 times to (hopefully) resolve chaining variable references.
```
saramarcondes, line 50: Why does this limitation exist? Why can't we just follow the chain until we reach a concrete value rather than a variable?

saramarcondes (approved):
LGTM and seems to work great. Left some nits and tips for types and some questions about the approach but otherwise 👍

packages/create-styles/src/createCompiler/plugins/cssVariables.js
ItsJonQ: Good call! I had no idea what it was. Turns out, it's represents `KEYFRAME`
https://github.com/thysultan/stylis.js/blob/4561e9bc830fccf1cb0e9e9838488b4d1d5cebf5/stylis.js#L121

packages/create-styles/src/cssCustomProperties/createRootStore.js
ItsJonQ: @saramarcondes Oooo!! I like this 👏

packages/create-styles/src/cssCustomProperties/createRootStore.js
ItsJonQ: Done!

packages/create-styles/src/cssCustomProperties/createRootStore.js
ItsJonQ: @saramarcondes I apologize... I don't know how to do this 🙈.

packages/create-styles/src/cssCustomProperties/createRootStore.js
ItsJonQ: @saramarcondes I gave this some thought! I think you're right 👍 .

I don't think there's a scenario where there may be stray / unresolvable CSS variable values.
The reason is because this resolver runs against the base G2 config values, which should alway compute to something.

It (currently) does not attempt to resolve custom theme values (which may contain unresolvable values).

Since this initial pass is to support base-level rendering for IE11, I think this is okay.

Thank you!!!

ItsJonQ:
@saramarcondes I pushed my fixes for the items you suggested (thank you!).
I'm not sure how to go about the typing stuff though 😊

saramarcondes (approved):
🚀 We can fix the types when we work on the whole of the `create-styles` package 👍

PR #97, opened 2020-11-09
Components: ExternalLink

saramarcondes:
Adds a new `ExternalLink` component in parity with `@wordpress/components/ExternalLink`.

### WP/Components
Captura de Tela 2020-11-09 às 09 47 11

### G2/Components
Captura de Tela 2020-11-11 às 09 44 29

### DIFF
--- a/packages/components/src/ExternalLink/ExternalLink.js
+++ b/packages/components/src/ExternalLink/ExternalLink.js
@@ -3,20 +3,16 @@ import { FiExternalLink } from '@wp-g2/icons';
 import React from 'react';
 
 import { Icon } from '../Icon';
+import { Link } from '../Link';
 import { VisuallyHidden } from '../VisuallyHidden';
 import * as styles from './ExternalLink.styles';
 
 function ExternalLink(props, forwardedRef) {
-	const { children, href, rel, ...otherProps } = useContextSystem(
-		props,
-		'ExternalLink',
-	);
+	const { children, ...otherProps } = useContextSystem(props, 'ExternalLink');
 
 	return (
-		<a
-			href={href}
+		<Link
 			ref={forwardedRef}
-			rel={rel}
 			// eslint-disable-next-line react/jsx-no-target-blank
 			target="_blank"
 			{...otherProps}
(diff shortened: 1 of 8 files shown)


packages/components/src/ExternalLink/ExternalLink.js
```diff
@@ -0,0 +1,36 @@
...
+import * as styles from './ExternalLink.styles';
+
+function ExternalLink(props, forwardedRef) {
+	const { children, href, rel, ...otherProps } = useContextSystem(
+		props,
+		'ExternalLink',
+	);
+
+	return (
+		<a
```
ItsJonQ, line 16: I think we can use the [`Link`](https://github.com/ItsJonQ/g2/tree/master/packages/components/src/Link) component here :D.

That should simplify prop destructuring a bit.

packages/components/src/ExternalLink/ExternalLink.styles.js
```diff
@@ -0,0 +1,6 @@
+import { css } from '@wp-g2/styles';
+
+export const Icon = css`
+	fill: currentColor;
```
ItsJonQ, line 4: @saramarcondes We can use the `color` prop for `<Icon />` to do:

```
<Icon color="currentColor" />
```

:)

packages/components/src/ExternalLink/ExternalLink.styles.js
```diff
@@ -0,0 +1,6 @@
...
+	margin-left: 0.1em;
```
ItsJonQ, line 5: I'm not too sure what `0.1em` would compute to 🤔 .

If it's something close to a spacing system value, would be be able to try the `ui.space()` util 🎉 
https://github.com/ItsJonQ/g2/blob/master/packages/styles/types/presets/space.d.ts#L14

ItsJonQ (approved):
@saramarcondes 🚀 from me! The last thing we gotta do is add it to the main `components/src/index.js` export 😊 .

Note: That file automatically creates a couple of snapshot tests for all exported components (with the exception of some manually excluded ones)

ItsJonQ (approved):
🚀 from me! Thanks @saramarcondes !!!

Issue #96, opened 2020-11-09
ButtonGroup: Add Component


ItsJonQ:
<img width="314" alt="Screen Shot 2020-11-09 at 10 26 45 AM" src="https://user-images.githubusercontent.com/2322354/98561243-a7b3dd00-2276-11eb-94af-044ddfe34450.png">

We need to add a `<ButtonGroup />` component to more easily handle UIs that involve clusters of Icon-based buttons (above). Clusters that need keyboard navigations and share a single "active" state.

At the moment, `SegmentedControl` does this, but it renders a different more bespoke UI.

ItsJonQ:
<img width="724" alt="Screen Shot 2020-11-09 at 11 52 27 AM" src="https://user-images.githubusercontent.com/2322354/98571899-041cf980-2283-11eb-99a5-922d4a3f9ac6.png">

Will have a PR soon! I'm adding some extra rendering features to `ButtonGroup` to enable different UIs :)

PR #98, opened 2020-11-09
Components: ButtonGroup

ItsJonQ:
Screen Shot 2020-11-09 at 11 52 27 AM

This update adds a new `ButtonGroup` component. The component can be used like this:

```jsx
<ButtonGroup onChange={setValue} value={value}>
  <Button icon={<FiBold />} value="bold" />
  <Button icon={<FiItalic />} value="italic" />
  <Button icon={<FiUnderline />} value="underline" />
</ButtonGroup>
```

Regular `Button` components are based into `ButtonGroup`, and they are automatically adjusted and rigged up via context.

##### `expanded`

Expanded makes the `ButtonGroup` go full width. The `Button` widths expand to (equally) take up the space within.

##### `segmented`

Segmented renders borders + splits between the inner `Button`.

`expanded` and `segmented` can be combined.

### 📻 RadioGroup

Under the hood, this component used the `Radio` components from Reakit for state management and interaction handling.

---

### 🕹 Testing

Check out the default `ButtonGroup` story. Ensure click the buttons can update the state. Make sure the other buttons in the `default` story update together (synced). This "sync" test ensures that the "controlled component" mechanic is correctly setup.

---

Resolves https://github.com/ItsJonQ/g2/issues/96


packages/components/src/Button/Button.js
```diff
@@ -11,14 +12,19 @@ function Button(props, forwardedRef) {
...
 		isControl = false,
 		isSubtle = false,
 		size = 'medium',
 		variant = 'secondary',
 		...otherProps
 	} = useContextSystem(props, 'Button');
 
+	const { buttonGroup } = useButtonGroupContext();
+	const isButtonGroupActive =
+		buttonGroup?.state && buttonGroup?.state === otherProps.value;
```
saramarcondes, line 25: Why the `&&`? Is `buttonGroup?.state === otherProps.value` not sufficient? I think they're _slightly_ different meanings but is the duplicated conditional get necessary?

saramarcondes (approved):
😻

packages/components/src/Button/Button.js
ItsJonQ: >  Is buttonGroup?.state === otherProps.value not sufficient?

It may be! I think I had issues in the beginning, but they may be unrelated. Checking now.

Thank you! ❤️

ItsJonQ: Update: Oh! I think it should be:

```jsx
	const isButtonGroupActive = buttonGroup && buttonGroup?.state === otherProps.value;
```

It **does** make a difference.

Here's where it's tricky.

`buttonGroup` may not exist. It only exists if the `Button` is used within `ButtonGroup`.
Cool.

Second part... `value` (initial value) may not be defined.

Which means that `buttonGroup?.state === otherProps.value` would evaluate to `true` for regular `Button` that live outside of `ButtonGroup`. (`undefined === undefined`)

To make this clearer, I can update the code to this:

```jsx
	const isWithinButtonGroup = !!buttonGroup;
	const isActiveWithinButtonGroup =
		isWithinButtonGroup && buttonGroup?.state === otherProps.value;
```

PR #99, opened 2020-11-10
JSDoc Types

saramarcondes:


sirreal:
I pushed a change so the build reports failures. What would we like to do with this? As far as I can tell there's a lot of work to get this mergeable. Maybe we should do this incrementally package-by-package?

saramarcondes:
@sirreal I think incrementally is the way to go. We can start with leaf (or almost leaf) packages like utils and substate.

saramarcondes:
Closing in favor of https://github.com/ItsJonQ/g2/pull/101

Issue #57, opened 2020-09-24
Optimizing render performance


ItsJonQ:
I'm noticing some sluggishness when there are many list entries with (complex) individual list items **and** if values need to rapidly update.

Might be able to improve render performance by optimizing how components like `Flex` enhance (clone) and render children. Perhaps something can be done with the Context mechanic (`connect` and `ComponentProvider`)

ItsJonQ:
Perhaps we can draw some inspiration from Reakit for context handling:
https://github.com/reakit/reakit/issues/745

ItsJonQ:
I just recorded a screencast where I share some ideas regarding performance as it relates to Design Tools + Gutenberg:

https://www.loom.com/share/c47612ce350b4584acd23493a44acf64

ItsJonQ:
Update!~ I just did a test that pushes interaction and cross-component prop/state interpolation. Results are very promising!

In other words... I did a test that made the UI update A BUNCH, and it's **fast**!!!!

## Demo

[**Video Demo**](https://d.pr/i/XnZ00n)
[**Video screencast + overview**](https://www.loom.com/share/e6cbb16ee80e47139a2478214714f592)

The demo simulates the data-binding relationships within Gutenberg. Specifically:

`wp-data` 👈 👉 Block controls
👇
Editor ("Canvas")

Larger components, basically any "Design Tool" as well as feature rich base level components, have to manage both inner state and sync incoming state (from props).

Achieving this is already quite challenging... even more so when rapid-fire updates need to be supported.

G2 Components accomplish this with mindful memoization and alternative state management techniques (via [Zustand](https://github.com/pmndrs/zustand))

The demo showcases these rapid-fire updates happening very smoothly without dropping frames.

Try it for yourself ✌️ :
https://g2-components.xyz/iframe.html?id=designtools-dataflow--default&viewMode=story

You can enable auto updating by adjusting the value under "Autopilot".

Drilling into some charts...

These are the performance results at level **11** speeds:

<img width="314" alt="Screen Shot 2020-10-07 at 2 34 42 PM" src="https://user-images.githubusercontent.com/2322354/95372948-4630d500-08aa-11eb-8b51-e8a42337ee24.png">

State consolidation + rendering happens comfortably within the 16.666ms windows.
This allows for the UI to render lag free 🎉

ItsJonQ:
Here's a fun update...

[...]

PR #100, opened 2020-11-11
Components: ClipboardButton

saramarcondes:
Add a new clipboard button that handles copying and a hook for when copying is finished.

No screenshots as it's just a button with some special props.

### DIFF
--- a/packages/components/src/ClipboardButton/ClipboardButton.js
+++ b/packages/components/src/ClipboardButton/ClipboardButton.js
@@ -1,16 +1,14 @@
 import { contextConnect, useContextSystem } from '@wp-g2/context';
 import { useClipboard } from '@wp-g2/utils';
 import { noop } from 'lodash';
-import React from 'react';
-import { useEffect } from 'react';
-import { useRef } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 
 import { Button } from '../Button';
 
 function ClipboardButton(props, forwardedRef) {
 	const {
 		text,
-		onClick: onClickProp = noop,
+		onClick = noop,
 		onCopy: onCopyProp = noop,
 		onFinishCopy = noop,
 		...otherProps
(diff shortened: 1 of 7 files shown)


packages/components/src/ClipboardButton/ClipboardButton.js
```diff
@@ -0,0 +1,37 @@
...
+		if (hasCopied) {
+			onCopyProp();
+		} else {
+			onFinishCopy();
+		}
+
+		lastHasCopied.current = hasCopied;
+	}, [hasCopied, onCopyProp, onFinishCopy]);
+
+	return <Button {...otherProps} onClick={onCopy} ref={forwardedRef} />;
```
ItsJonQ, line 34: @saramarcondes Would love your thoughts on this. I think we should pass-through an `onClick` callback as well (from props), in addition to firing the `onCopy` callback. I'd expect `onClick` to work (as a consumer), even though under the hood it's firing `onCopy`. What do you think :)

saramarcondes: I had the same thought, I'm glad you brought it up. I think it makes sense to merge the two as you suggested 👍

packages/components/src/ClipboardButton/ClipboardButton.js
```diff
@@ -0,0 +1,45 @@
...
+import React from 'react';
+import { useEffect } from 'react';
+import { useRef } from 'react';
+
+import { Button } from '../Button';
+
+function ClipboardButton(props, forwardedRef) {
+	const {
+		text,
+		onClick: onClickProp = noop,
```
ItsJonQ, line 13: I don't think we need to alias `onClick` to `onClickProp` here :)
(unlike `onCopy`, which comes from `useClipboard`)

packages/components/src/ClipboardButton/ClipboardButton.js
```diff
@@ -0,0 +1,45 @@
...
+		if (hasCopied) {
+			onCopyProp();
+		} else {
+			onFinishCopy();
+		}
+
+		lastHasCopied.current = hasCopied;
+	}, [hasCopied, onCopyProp, onFinishCopy]);
+
+	const handleOnClick = (event) => {
```
ItsJonQ, line 35: @saramarcondes I've been memoizing callbacks G2 Components. I didn't originally... how important it was for [performance](https://github.com/ItsJonQ/g2/issues/57) 🙈 (assuming the consumer also passes in a memoized callback).

It's slightly tedious, but I think the trade off and value is worth it 🤞

ItsJonQ:
@saramarcondes Haii! I left some comments :).

The last thing we gotta do is export this component in the main `src/index.js`.

Unfortunately, the `remake` command doesn't automatically add it in the `src/index.js` file 😊

ItsJonQ (approved):
🚀 from me! Thanks @saramarcondes !!!

PR #104, opened 2020-11-13
Types: create-styles and styles

saramarcondes:
This PR adds types to create-styles and styles.

Proof it works 😎 

Captura de Tela 2020-11-13 às 15 31 17
Captura de Tela 2020-11-13 às 15 31 02

These screenshots are taken from inside the `styles` package.

### Testing instructions

* Manually check generated types to make sure they look right (I mean skim them probably).
* Consume the types from `styles` and pay special attention to `styled` and `get` which are the most complicated types we generate. Make sure they continue to work as expected with the appropriate type hints 😎 
* Make sure the components still work 😱


packages/create-styles/src/createStyleSystem/createStyledComponents.js
```diff
@@ -1,13 +1,154 @@
...
+import { getDisplayName, hoistNonReactStatics, is } from '@wp-g2/utils';
 import React from 'react';
 
 import { tags } from './tags';
-import { compileInterpolatedStyles, getDisplayName } from './utils';
+import { compileInterpolatedStyles } from './utils';
+
+/**
+ * @typedef {{
+	(component: import('react').ReactElement | string): import('./createCoreElement').CreatePolymorphicComponent<{}, 'div'>;
```
ItsJonQ, line 9: Holy moly! 😮

ItsJonQ:
Excited for this! Although, I have no idea what's going on in the code 😂

saramarcondes:
I actually want to call out that intellisense is actually smarter after this PR:

Captura de Tela 2020-11-13 às 15 35 27

ItsJonQ:
> I actually want to call out that intellisense is actually smarter after this PR
@saramarcondes Heck yes! 🤘

packages/create-styles/src/createStyleSystem/tags.js
```diff
@@ -66,7 +67,6 @@ export const tags = [
 	'main',
 	'map',
 	'mark',
-	'marquee',
```
saramarcondes, line 69: `marquee` is deprecated and not supported by React. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L3052

ItsJonQ: Ah, I vaguely remember this from emotion. Feel free to remove or comment out 👍

packages/utils/src/is.js
```diff
@@ -37,6 +37,8 @@ export const is = {
 	file: (o) => o instanceof File,
 	numeric,
 	numericZero,
+	/** @type {(o: any) => o is import('create-emotion').ObjectInterpolation} */
+	objectInterpolation: (o) => _.isPlainObject(o),
```
saramarcondes, line 41: This is kind of a hack? What do you think @sirreal?

packages/utils/src/math/interpolate.js
```diff
@@ -74,7 +75,7 @@ export function interpolate(
 	if (isColor(outputRange1) && isColor(outputRange2)) {
 		const mixAmount = interpolate(input, inputRange, [0, 100]);
 		return colorize
-			.mix(outputRange1, outputRange2, mixAmount)
+			.mix(outputRange1.toString(), outputRange2.toString(), mixAmount)
```
saramarcondes, line 78: The `ColorInput` type only accepts strings, not numbers (even though it works fine if you pass a number). It might be worth updating upstream?

packages/utils/src/strings.js
```diff
@@ -1,17 +1,14 @@
-export { default as kebabCase } from 'lodash.kebabcase';
-export { default as capitalize } from 'lodash.capitalize';
-export { default as upperFirst } from 'lodash.upperfirst';
+import _ from 'lodash';
+
+export const kebabCase = _.kebabCase;
+export const capitalize = _.capitalize;
+export const upperFirst = _.upperFirst;
+export const repeat = _.repeat;
```
saramarcondes, line 6: Using the `lodash.whatever` exports (a) doesn't get us anything and (b) prevents lodash's types from filtering through as the types are only defined for the actual `lodash` package.

packages/utils/src/strings.js
```diff
@@ -1,17 +1,14 @@
...
- * @param {string} char Character to repeat.
- * @param {number} n Number of times to repeat.
- * @return {string} String with repeated characters
+ * @param {string} str
+ * @return {string}
  */
-export function repeat(char, n, a) {
-	return (a = []).join((a[n - 1] = char));
-}
-
```
saramarcondes, line 14: Not sure what was going on here but `lodash` has it's own `repeat` that wil work fine for all the usages of this function.

sirreal (commented):
Nice work! There's a lot here, it's pretty tricky to review without going through the process of typing again 🙂 

@saramarcondes I see you've started calling out some things you have questions about, that makes it much easier for me. Otherwise, I'd like to trust `tsc` that things are sound and can adjust if we don't get things right.

packages/utils/src/is.js
sirreal: Yeah… I assume this was to satisfy the compiler? I'd rather use a careful assertion than export a new, relatively weak "is" funciton 🙂

packages/utils/src/math/interpolate.js
sirreal: Maybe. `Infinity`, `NaN` and others are in the `Number` type. You can certainly pass bad strings in the string type as well although [template literal types in TypeScript 4.1](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#template-literal-types) may allow some interesting types.

packages/utils/src/strings.js
sirreal: I was going to comment on this 🙂  I'd like to stop re-exporting these completely and use them directly as needed.

sirreal: Middle ground would be like `export { kebabCase } from 'lodash'`, but that will probably involve fiddling with `allowSyntheticDefaultImports` or other config.

packages/components/src/Card/__tests__/__snapshots__/Card.test.js.snap
```diff
@@ -150,7 +150,7 @@ exports[`props should render elevation 1`] = `
     data-g2-component="CardElevation"
   />
   <div
-    class="components-elevation wp-components-elevation css-fmvmxl"
+    class="components-elevation wp-components-elevation css-zytxex"
```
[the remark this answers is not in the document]
ItsJonQ: @sirreal #111 is merged! 🎉

packages/components/src/AlignmentMatrixControl/__tests__/AlignmentMatrixControl.test.js
```diff
@@ -6,7 +6,7 @@ import { AlignmentMatrixControl } from '../index';
 describe('props', () => {
 	test('should render correctly', () => {
 		const { container } = render(
-			<AlignmentMatrixControl value="center center" />
+			<AlignmentMatrixControl value="center center" />,
```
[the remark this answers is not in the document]
ItsJonQ: @sirreal G2 is using regular ol' [`prettier`](https://github.com/ItsJonQ/zero/blob/master/package.json#L113), coming from the [`@itsjonq/zero` build scripts](https://github.com/itsjonq/zero).

It should smooth itself out when we move the code over to the Gutenberg repo 👍 

We'll just do a massive `prettier` lint+fix 🤞

packages/utils/src/strings.js
ItsJonQ: @saramarcondes Let's use lodash's repeat!!

packages/utils/src/strings.js
ItsJonQ: > I'd like to stop re-exporting these completely and use them directly as needed

Just to confirm! For example, a file within the components package, would we do:

```
import { kebabCase } from 'lodash'
```

---

For some context, I wasn't sure what the (long term) stance was for `lodash`, which is why I initially kept it's usage minimal with the G2 Components development.

If we're certain lodash will be with us for a while, then let's just use it :).

ItsJonQ:
@saramarcondes Haiii! I merged in @sirreal 's [jest/emotion update](https://github.com/ItsJonQ/g2/pull/111). You may need to merge in the latest `master` to resolve snapshot conflicts 🙈

packages/utils/src/strings.js
saramarcondes: > Middle ground would be like export { kebabCase } from 'lodash', but that will probably involve fiddling with allowSyntheticDefaultImports or other config.

Correct, that doesn't work with the current config 🤔 

> Just to confirm! For example, a file within the components package, would we do:

Yes, that is preferred to re-exporting everything.

> If we're certain lodash will be with us for a while, then let's just use it :).

Given `lodash` is used extensively in Gutenberg it's probably safe to say it's going to stick around 🙂 🙃

packages/utils/src/is.js
saramarcondes: I'd rather that too, but I could not figure out what assertion to use 🤔 `ObjectInterpolation` is basically a special `Record` with specific keys, right?

```
{
  [K in keyof CSSProperties]:
    | CSSProperties[K]
    | Array<Extract<CSSProperties[K], string>>
}
```

And because `Interpolation` also includes `string` `number` and `null` `undefined` you have to do a bunch of assertions to get here. Any ideas?

packages/create-styles/src/components/ThemeProvider/ThemeProvider.js
```diff
@@ -18,14 +18,19 @@ import {
...
+ * @property {import('../../createStyleSystem/generateTheme').GenerateThemeResults} globalStyles Styles to apply globally.
  * @property {boolean} isDark Determines if dark-mode styles should be rendered.
  * @property {boolean} isColorBlind Determines if color-blind-mode styles should be rendered.
  * @property {boolean} isReducedMotion Determines if reduced-motion-mode styles should be rendered.
  * @property {boolean} isHighContrast Determines if high-contrast-mode styles should be rendered.
- * @property {object} theme Custom theme properties.
+ * @property {Record<string, string>} theme Custom theme properties.
+ * @property {Record<string, string>} darkTheme Custom dark theme properties.
+ * @property {Record<string, string>} highContrastTheme Custom high contrast theme properties.
+ * @property {Record<string, string>} darkHighContrastTheme Custom dark & high contrast theme properties.
```
saramarcondes, line 33: I think this should be `StyleConfig`

packages/create-styles/src/components/ThemeProvider/ThemeProvider.utils.js
```diff
@@ -108,7 +108,7 @@ export function useReducedMotionMode({
 	isReducedMotion,
 	ref,
 }) {
-	const [, setIsReducedMotion] = useReducedMotion(isReducedMotion);
+	const [, setIsReducedMotion] = useReducedMotion();
```
saramarcondes, line 111: `useReducedMotion` never accepted any parameters.

packages/create-styles/src/components/ThemeProvider/ThemeProvider.utils.js
```diff
@@ -150,13 +150,16 @@ export function useThemeStylesStore() {
 
 /**
  * @typedef UseThemeStyles
+ * @property {import('create-emotion').Emotion['injectGlobal']} injectGlobal
  * @property {boolean} isGlobal Determines if the theme styles are rendered globally or scoped locally.
- * @property {object} theme Custom theme values.
+ * @property {Record<string, string>} theme Custom theme values.
```
saramarcondes, line 155: `StyleConfig`

packages/create-styles/src/components/ThemeProvider/ThemeProvider.utils.js
```diff
@@ -199,8 +203,10 @@ export function useThemeStyles({
 		 * We only want to update + set the theme if there's a change.
 		 * Since themes (potentially) be nested, we need to do a shallowEqual check.
 		 */
+		// @ts-ignore
 		if (shallowEqual(themeRef.current, theme)) return;
 
+		// @ts-ignore
 		themeRef.current = theme;
```
saramarcondes, line 210: Not allowed to assign to `current` as it is `readonly` 🤔 Is this something we just ignore when writing libraries?

packages/utils/src/is.js
saramarcondes: I think I've got it... we can make this safe by only allowing `TemplateStringsArray | Interpolation` to be passed in.

packages/styles/src/css.js
```diff
@@ -123,25 +125,25 @@ export const responsive = (styles = {}) => {
...
 			}
 		}
-		return compile(...[arg, ...rest]);
+		return compile(template, ...args);
 	}
 
-	return compile(...args);
+	// @ts-ignore
+	return compile(template, ...args);
 }
```
saramarcondes, line 149: @ItsJonQ Can you confirm this still matches your intention or am I missing something? In particular I'm concerned about `is.objectInterpolation` whether that was exactly your intention here or if there was something else that might get passed in your were looking for.

saramarcondes:
I need to take a break from this today to work on some other stuff but I will return to it tomorrow.

So far I think there's a couple things I want to do with this PR:

1. Get the build passing again
2. Ensure all `// @ts-ignore` directives have an explanation with them
3. Address all `@todo`s I left dangling around.

ItsJonQ:
Thank you all for all of your fantastic work 🙏 .

I just took this repo for a spin. There seems to be something up with styling compiling 🤔 

`master`:
Screen Shot 2020-11-16 at 12 07 57

This branch:
Screen Shot 2020-11-16 at 12 08 45

You can check it out in the Accordion story in the deploy preview:
https://g2-git-try-type-create-styles.itsjonq.vercel.app/?path=/story/components-accordion--default

**I'm not sure what may be causing it**.

I tried reverting the changes in `packages/styles/src/css.js`, but no luck

🤔

ItsJonQ:
☝️ Update! I don't know why it's breaking (yet). This may help.

Check out compiled styles for this PR:

> Screen Shot 2020-11-16 at 12 32 44 PM

The extraSpecificity plugin doesn't seem to be kicking in.

---

**That aside**... the style should still be working. Order shouldn't matter.

---

I think this PR exposed a bug in the code!

After some research, it appears that the `*` selector (universal selector) has a [specificity score of `0`](https://www.smashingmagazine.com/2007/07/css-specificity-things-you-should-know/).

I guess I have to figure out another way to for `Flex` to target it's children 🤔

packages/styles/src/css.js
ItsJonQ: This looks correct! Thank you @saramarcondes ❤️

saramarcondes:
@itsjonq Thanks for catching that bug! It turned out to be a simple misplaced `else` here: https://github.com/ItsJonQ/g2/pull/104/commits/1819acc259fb2dfdd81d74d1b68044e063df6b1e

packages/create-styles/src/components/ThemeProvider/ThemeProvider.utils.js
sirreal: There's a note about this here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/77d6faf16f14832e88c28aa34bf86e3a67520a69/types/react/index.d.ts#L1029-L1043

We'll want a `MutableRefObject` [above](https://github.com/ItsJonQ/g2/pull/104/files/e487d30475a9d141405ddfae473b949cd497693a#diff-b132a18aa8ec54f906c1b8484058cd5d3f636def3bdfe93ac37a58bc16b511efR176).

ItsJonQ:
@saramarcondes Haiii! I hope you don't mind, I fixed the some tests regarding the `responsive` utility :).

packages/styles/src/presets/colors.js
```diff
@@ -4,9 +4,149 @@ import { get } from '../core';
...
+	// The system WordPress blue color.
+	wordpressBlue: string;
+	// The system white color.
+	white: string;
+ }} ColorInstance
+ */
+
+/** @type {ColorInstance} */
+// @ts-ignore We add the missing properties below.
+const color = (value) => {
```
saramarcondes, line 147: @sirreal Do you know if there's a way around this `ts-ignore`?

packages/create-styles/src/createStyleSystem/createStyleSystem.js
```diff
@@ -134,6 +153,7 @@ export function createStyleSystem(options = defaultOptions) {
 		rootStore,
 	};
 
+	// @ts-ignore `get` is problematic and difficult to get right here... maybe warrants an @todo
```
saramarcondes, line 156: @sirreal Would love some help on this one, I'm having a hard time parsing through the error message coming out of this (and an even harder time trying to figure out how to fix it). I'm like 99.99999999% certain what I've done is sound but something's not adding up.

sirreal:
Thanks @saramarcondes, nice work 👏

PR #111, opened 2020-11-14
Add: @emotion/jest

sirreal:
These snapshots look more stable and informative 🙂 

We can probably go and do the same in Gutenberg.

I also deleted some stray package-lock files that seem to be leftovers. The project is using yarn, right?


ItsJonQ:
> I also deleted some stray package-lock files that seem to be leftovers. The project is using **yarn**, right?

Whoops! 🙈 . Yes indeed, this project uses yarn :)

ItsJonQ (approved):
@sirreal Thank you for this! It's exciting to see the compiled styles alongside the markup in the snapshots - especially since we're doing some interesting things in the preprocessor's plugin layer.

Good timing as well! I just created an issue to potentially update G2 to Emotion v11, which will involve some upgrades in the `create-styles` package.

https://github.com/ItsJonQ/g2/issues/112

Having these updated snapshots could help a lot with this!

Issue #119, opened 2020-11-16
Flex: Fix child selector specificity


ItsJonQ:
The issue is the `> * + *` based rule isn't specific enough to target children that may have `margin: 0` rules.

This is because the CSS universal selector (`*`) has a specificity rating of zero. Compounding it like `* + *` doesn't change this.

As a work around... we can do something like this:

```css
> * + *:not(marquee) {
}
```

The `:not()` will bump it's specificity rating. And there's a 99.9999999999999999999% guarantee that the element would not be a `marquee` within the context of React + modern web UI.

Here's a screenshot of this solution working:

<img width="426" alt="Screen Shot 2020-11-16 at 12 44 44 PM" src="https://user-images.githubusercontent.com/2322354/99288804-24603180-280a-11eb-8f76-1df248eb3139.png">

---

This bug was originally spotted here:
https://github.com/ItsJonQ/g2/pull/104#issuecomment-728216780

PR #120, opened 2020-11-16
Flex: Fix style rendering for child items

ItsJonQ:
The issue is the `> * + *` based rule isn't specific enough to target children that may have `margin: 0` rules.

This is because the CSS universal selector (`*`) has a specificity rating of zero. Compounding it like `* + *` doesn't change this.

As a work around... we can do something like this:

```css
> * + *:not(marquee) {
}
```

The `:not()` will bump it's specificity rating. And there's a 99.9999999999999999999% guarantee that the element would not be a `marquee` within the context of React + modern web UI.

Here's a screenshot of this solution working:

Screen Shot 2020-11-16 at 12 44 44 PM

---

Resolves:
https://github.com/ItsJonQ/g2/issues/119


ItsJonQ:
Will merge this up 😊

Issue #109, opened 2020-11-14
UnitInput/Slider: Need better state management solution to sync controlled state


ItsJonQ:
<img width="189" alt="Screen Shot 2020-11-14 at 11 35 39 AM" src="https://user-images.githubusercontent.com/2322354/99152081-a1a26f80-266d-11eb-917b-af1a089af630.png">

At the moment, a `UnitControl` can be composed with a `Slider` like so:

```jsx
<Grid>
	<UnitControl value={value} onChange={onChange} />
	<Slider value={value} onChange={onChange} />
</Grid>
```

This is okay for purely numeric based values. It gets tricky when units are involved.

This is because the `Slider` does not provided a unit value through it's `onChange` callback.
It does, however, [parsed the incoming value](https://github.com/ItsJonQ/g2/blob/master/packages/components/src/Slider/useSlider.js#L31) to strip potential units.

---

### 💡 Idea

One update we can do is to make `Slider` work with unit values automatically.

It would probably work something like this:

1. Parse incoming value
2. Store potential `unit`
3. Use incoming value numeric value as controlled state
4. `onChange` -> if there's a `unit`, combine the next value + unit together. Otherwise, pass along the numeric value.

---

Sidenote: This built in unit support may be something we want to add in future control types (e.g. an [`AngleInput`](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/angle-picker-control))

PR #122, opened 2020-11-16
Slider: Add support for unit handling

ItsJonQ:
This update adds built-in unit handling support for `Slider`. `Slider` now knows how to understand incoming unit values (e.g. `50px`). It also knows how to preserve unit values during an `onChange` event.

For example:

```jsx
const [value, setValue] = useState('50px')

<Slider value={value} onChange={setValue} />
```

The `Slider` will parse and pass the value of `50` to the internal `input`.

During a change, say to a value of `49`, the `onChange` callback will be `49px`. This is because `Slider` accounts for a potential unit passed in via the `value` prop.

`Slider` also knows how to account for situations where the value may change 👍 

Resolves https://github.com/ItsJonQ/g2/issues/109

### DIFF
--- a/packages/design-tools/src/__stories__/Presentation/shared.js
+++ b/packages/design-tools/src/__stories__/Presentation/shared.js
@@ -26,7 +26,6 @@ import { ui } from '@wp-g2/styles';
 import { createStore, shallowCompare } from '@wp-g2/substate';
 import { is } from '@wp-g2/utils';
 import React from 'react';
-import CSSUnit from 'units-css';
 
 export const Wrapper = ({ children, title }) => {
 	return (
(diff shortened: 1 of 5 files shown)


packages/design-tools/src/__stories__/Presentation/shared.js
```diff
@@ -277,31 +276,15 @@ export const CombinedFormGroupInputSlider = React.memo(
 		const value = useTypography((state) => state[prop], shallowCompare);
 
 		const handleOnChange = React.useCallback(
-			(value) => {
-				typographyStore.setState((prev) => {
-					// Handles unit changes
-					const unit =
```
ItsJonQ, line 283: It's so simple now 😍

PR #125, opened 2020-11-17
Resolve @wp-g2/a11y to upstream reakit

saramarcondes:
Updates all references to the a11y package to be directly to reakit and removes the a11y package.


ItsJonQ (approved):
🚀 !!! Thanks @saramarcondes !!!

PR #127, opened 2020-11-17
Resolve @wp-g2/gestures to upstream react-use-gesture

saramarcondes:
Changes references to the `gestures` packages to `react-use-gesture` proper and removes the `gestures` package.


ItsJonQ (approved):
🚀 from me!!!!

PR #133, opened 2020-11-17
Replace wp-g2/icons with wordpress/icons when possible

saramarcondes:


packages/components/src/ButtonGroup/__stories__/ButtonGroup.stories.js
```diff
@@ -1,4 +1,4 @@
-import { FiBold, FiItalic, FiUnderline } from '@wp-g2/icons';
+import { formatBold, formatItalic, formatUnderline } from '@wordpress/icons';
```
saramarcondes, line 1: `formatUnderline` exists as of 2 weeks ago and should be available in the next release. Though currently it renders nothing.

packages/components/src/Icon/Icon.js
```diff
@@ -10,6 +10,7 @@ function Icon(props, forwardedRef) {
 		children,
 		className,
 		color,
+		fill = color,
```
saramarcondes, line 13: `@wordpress/icons` require `fill` rather than `color`. This may be a hack, not sure, will leave final approval for this solution up to @ItsJonQ

packages/components/src/ButtonGroup/__stories__/ButtonGroup.stories.js
ItsJonQ: That's alright :). It's not too important that they work for the stories.

packages/components/src/Icon/Icon.js
ItsJonQ: Hmm 🤔 . This one is an interesting one. Unlike other components (like `Button`), Gutenberg's `Icon` comes from a non `@wordpress/components` package.

For a second, I thought, "Well... what if we just used `Icon` from `@wordpress/icons` instead".
But then I realized that the G2 `Icon` component has many advantages, such as it being powered by `View` and it being tapped into the Style + Context system.

Ultimately, because we're using `@wordpress/icons` **svg** code, we have to accommodate the wrapper (that being `Icon`) for that code. For example, [react-icons](https://github.com/react-icons/react-icons) SVGs are built to use `color`.

Based on this, I think we should go with `fill`.

If we do, we can remove the `color` prop from this `Icon` component.

@saramarcondes Lemme know what you think ❤️

saramarcondes: Sounds good to me!

ItsJonQ:
@saramarcondes Oops! I think I fumbled up the components `package.json` with [this PR](https://github.com/ItsJonQ/g2/pull/134) during merging. I accidentally removed `@wordpress/i18n` dependency 🙈 

Sorry!

ItsJonQ:
Oh lol.

The `gif diff` wrapper code was committed 🙈 

Screen Shot 2020-11-17 at 5 00 46 PM

packages/components/src/ExternalLink/ExternalLink.js
```diff
@@ -1,6 +1,9 @@
+<<<<<<< HEAD
```
ItsJonQ, line 1: Whoopsies 😊 . Need to remove these, haha

ItsJonQ (approved):
🚀 from me!!! Thank you @saramarcondes !!!!

packages/components/src/BaseButton/BaseButton.styles.js
```diff
@@ -155,12 +157,14 @@ export const LoadingOverlay = css`
 export const subtle = css`
 	border-color: ${ui.get('controlBorderColorSubtle')};
 	color: ${ui.color.text};
+	fill: ${ui.color.text};
```
ItsJonQ, line 160: I wish we didn't have to do this 🙈 
Maybe an enhancement we can make in the future

Issue #136, opened 2020-11-17
Icon: Improve "fill" color


ItsJonQ:
This idea came from the recent `Icon` component + package refactor,
https://github.com/ItsJonQ/g2/pull/133/files/eb4a8c7ff3e643569bd26a2f62516c9ea99258e3#r525559793

Because `@wordpress/components` uses `fill` for it's Icon colors (vs. `color` from `react-icons`), we needed to add the `fill` CSS property to various styles (e.g. `BaseButton`) to accommodate this.

### Problem

I can imagine this not scaling well, as components (G2 or not) may need to adjust icon color from a higher node vs. direct manipulation through `<Icon fill="red" />`. In the case of `Button`, by default, we want the `Icon` to absorb the (text) color from the `Button`.

### Solution

I think we can do this for `Icon.js`

```jsx
		sx.fill = css({
			color: fill,
			fill: 'currentColor',
		});
```

We're still passing in the `fill` prop into `<Icon fill="red" />`. However, the `Icon` will use that for the `color` CSS prop instead, and `fill` will absorb it through `currentColor`.

Here's a quick test of this solution working:

<img width="490" alt="Screen Shot 2020-11-17 at 5 24 00 PM" src="https://user-images.githubusercontent.com/2322354/99457953-c7dc3f80-28f9-11eb-831b-9c0b74e58137.png">

I've removed the recently added `fill` styles in `BaseButton` and used the `currentColor` strategy above.
We can see that the `X` close icons are correctly absorbing the colors for the `Alert`.

PR #137, opened 2020-11-17
Icon: Improve "fill" color rendering

ItsJonQ:
This update improves how the `Icon` component handles the `fill` rendering for `@wordpress/icons` svgs. 

This idea came from the recent `Icon` component + package refactor,
https://github.com/ItsJonQ/g2/pull/133/files/eb4a8c7ff3e643569bd26a2f62516c9ea99258e3#r525559793

Because `@wordpress/components` uses `fill` for it's Icon colors (vs. `color` from `react-icons`), we needed to add the `fill` CSS property to various styles (e.g. `BaseButton`) to accommodate this.

### Problem

I can imagine this not scaling well, as components (G2 or not) may need to adjust icon color from a higher node vs. direct manipulation through `<Icon fill="red" />`. In the case of `Button`, by default, we want the `Icon` to absorb the (text) color from the `Button`.

### Solution

I think we can do this for `Icon.js`

```jsx
		sx.fill = css({
			color: fill,
			fill: 'currentColor',
		});
```

We're still passing in the `fill` prop into `<Icon fill="red" />`. However, the `Icon` will use that for the `color` CSS prop instead, and `fill` will absorb it through `currentColor`.

Here's a quick test of this solution working:

Screen Shot 2020-11-17 at 5 24 00 PM

I've removed the recently added `fill` styles in `BaseButton` and used the `currentColor` strategy above.
We can see that the `X` close icons are correctly absorbing the colors for the `Alert`.

---

Resolves https://github.com/ItsJonQ/g2/issues/136

### DIFF
--- a/packages/components/src/Alert/__tests__/__snapshots__/Alert.test.js.snap
+++ b/packages/components/src/Alert/__tests__/__snapshots__/Alert.test.js.snap
@@ -245,8 +245,6 @@ exports[`props should render as dismissable 1`] = `
   display: -webkit-inline-flex;
   display: -ms-inline-flexbox;
   display: inline-flex;
-  fill: #1e1e1e;
-  fill: var(--wp-g2-color-text);
   font-size: 13px;
   font-size: var(--wp-g2-font-size);
   line-height: 1;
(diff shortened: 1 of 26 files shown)


ItsJonQ:
Going to merge this up! 🙏

PR #157, opened 2020-11-23
Use is-shallow-equal

### LINKED ISSUE #141: Utils: Use `@wordpress/is-shallow-equal`
Currently we use a version of `shallowEqual` taken from `reakit-utils`. We should just use the `@wordpress/is-shallow-equal` package directly.

saramarcondes:
Use `@wordpress/is-shallow-equal` instead of a new function.

* Uses the utility from WordPress core
* Removes the one written in G2

### Testing instructions

Ensure storybook loads and that switching themes still works.

Closes #141


ItsJonQ (approved):
🚀 from me!! Thank you @saramarcondes

PR #160, opened 2020-11-23
Rename components A through Card to be kebab-cased

saramarcondes:


ItsJonQ:
Chatted with @saramarcondes .
Closing this. We'll rename them when we move over 🙏

PR #165, opened 2020-11-24
Refactor PolymorphicComponent type and type context

saramarcondes:


packages/components/src/BlankSlate/BlankSlate.js
```diff
@@ -10,6 +10,17 @@ import { Text } from '../Text';
 import { VStack } from '../VStack';
 import * as styles from './BlankSlate.styles';
 
+/**
```
saramarcondes, line 13: This serves as an example of how we'll type components in the future once we move their types out of the `.d.ts` files.

packages/components/src/HStack/__tests__/__snapshots__/HStack.test.js.snap
```diff
@@ -90,15 +90,8 @@ exports[`props should render Spacer 1`] = `
...
-  display: block;
-  max-height: 100%;
-  max-width: 100%;
-  min-height: 0;
-  min-width: 0;
-  display: var(--wp-g2-flex-item-display);
-  -webkit-flex: 1;
-  -ms-flex: 1;
-  flex: 1;
+  margin-bottom: calc(4px * 2);
```
saramarcondes, line 93: This was a bug fix 🙂 Previously `hasNamespace` did not work correctly 😱 This snapshot change is evidence that it is now working, where spacers are receiving only their styles and not the flex item styles.

packages/components/types/components/Alert.d.ts
```diff
@@ -41,7 +41,7 @@ export declare type AlertProps = {
  * </Alert>
  * ```
  */
-export declare const Alert: PolymorphicComponent<AlertProps>;
+export declare const Alert: PolymorphicComponent<'div', AlertProps>;
```
saramarcondes, line 44: This parameter order changed and it is now intentionally required (to avoid mistakes). This is clearer when we're typing the component next to the code rather than in definition files as it'll be more obvious what the box is for each component.

packages/create-styles/src/create-style-system/polymorphic-component.ts
```diff
@@ -0,0 +1,43 @@
...
+	[P in keyof JSX.IntrinsicElements]: PolymorphicComponent<P, {}>;
+};
+
+type CreateStyledComponents = {
+	[P in keyof JSX.IntrinsicElements]: CreatePolymorphicComponent<P, {}>;
+};
+
+export type CreateStyled = CreateStyledComponents & {
+	<T extends As>(component: T): CreatePolymorphicComponent<T, {}>;
+};
```
saramarcondes, line 43: I'm particularly proud of these types, where we get strongly typed core elements and styled components in a single line 😎

ItsJonQ: Holy. Smokes.

❤️ @saramarcondes ❤️

ItsJonQ (approved):
🚀 from me @saramarcondes !!! This PR is phenomenal!!! 😍

PR #167, opened 2020-11-24
Type Accordion

saramarcondes:


packages/components/src/Accordion/useAccordion.js
```diff
@@ -22,20 +65,23 @@ export function useAccordion(props) {
 		...otherProps
 	} = useContextSystem(props, 'Accordion');
 
-	const accordionStore = useSubState((set) => ({
+	/** @type {import('zustand').UseStore<AccordionStore>} */
+	const accordionStore = useSubState((
```
ItsJonQ, line 69: Interesting! This example is super helpful. It'll help me understand how to type (or try to type) components using `substate` 🎉

packages/components/src/Accordion/useAccordion.js
```diff
@@ -65,51 +111,54 @@ export function useAccordion(props) {
...
-		() => ({
-			// Actions
-			update: ({ id, visible = false }) => {
-				if (visible) {
-					if (allowMultiple) {
-						accordionStore.getState().add(id);
-					} else {
-						accordionStore.getState().set(id);
-					}
+	/** @type {import('zustand').UseStore<Store>} */
```
ItsJonQ, line 122: I guess we can't do `import('@wp-g2/substate')` for this?

packages/substate/src/use-sub-state.js
```diff
@@ -2,9 +2,9 @@ import { useRef } from 'react';
 import createStore from 'zustand';
 
 /**
- *
- * @param  {Parameters<typeof createStore>} args
+ * @template {Record<string | number | symbol, unknown>} TState
```
ItsJonQ, line 5: Holy smokes 🙈 👍

ItsJonQ:
@saramarcondes 👍 from me! I just had a question about one of the types:

> // packages/components/src/Accordion/useAccordion.js
> /** @type {import('zustand').UseStore<Store>} */

Are we not able to import `@wp-g2/substate`? (and `substate` and "infer" `zustand`, if that's even the right word for it, haha)

ItsJonQ (approved):
🚀

saramarcondes:
> Are we not able to import @wp-g2/substate? (and substate and "infer" zustand, if that's even the right word for it, haha)

Unfortunately to do that we would have to re-export all of the types from zustand (the same way that `@wordpress/element` does for React's types). I don't see any value in maintaining that and there's nothing lost as far as I can tell by importing directly from `zustand`. Even in `@wordpress/components` we import from `react` rather than `@wordpress/element` for React's types.

PR #169, opened 2020-11-25
Type ExternalLink and its dependencies

saramarcondes:
Types:

* Truncate
* Text
* Link
* ExternalLink
* Icon


packages/components/src/Text/types.ts
```diff
@@ -0,0 +1,188 @@
+import { Props as TruncateProps } from '../Truncate/useTruncate';
```
saramarcondes, line 1: @ItsJonQ This is what I was talking about using `.ts` files to define the types that we then import into the JSDoc 😉

packages/components/src/Text/useText.js
```diff
@@ -128,7 +132,16 @@ export function useText(props) {
 	 */
 	if (!truncate && is.array(children)) {
 		content = React.Children.map(children, (child) => {
-			if (!is.plainObject(child)) return child;
+			// Thiis is... unfortunate. Would we rather throw a `ts-ignore` here than increase the complexity of this loop?
```
saramarcondes, line 135: /cc @ItsJonQ @sirreal

packages/components/src/Text/types.ts
ItsJonQ: Thank goodness for that!~!!

packages/components/src/Text/useText.js
ItsJonQ: I mean. I'm okay with that 😂

ItsJonQ:
@saramarcondes This is phenomenal!!! I feel like I'm able to type things given your examples - at least, at the `components` package level. (Goodness, I can't even at the `create-styles` or `context` level).

It's wonderful to see the dedicated `types.ts` file working 🙌 .
1. For the inline examples
2. Because it's easier to read/write (vs. JSDocs 😂 )

packages/components/src/Text/useText.js
saramarcondes: Yeah I think I'll just add some surgical ts-ignores. Otherwise this is too complex for a loop

PR #170, opened 2020-11-25
Type Button and its dependencies (and Spinner)

saramarcondes:
Types:

* Spinner (cause what the heck why not another good simple example)
* ControlGroup
* ButtonGroup
* BaseButton
* Button


packages/components/src/Button/Button.js
```diff
@@ -35,11 +52,12 @@ function Button(props, forwardedRef) {
...
 				isWithinButtonGroup &&
 				enableSelectNone &&
-				isButtonGroupActive
+				isButtonGroupActive &&
+				buttonGroup
 			) {
 				event.preventDefault();
 				event.stopPropagation();
-				buttonGroup.setState(null);
+				buttonGroup.setState(undefined);
```
saramarcondes, line 60: `setState` actually doesn't accept `null` but it does accept `undefined` 🙂

packages/components/src/ButtonGroup/ButtonGroup.Context.js
```diff
@@ -1,4 +1,7 @@
 import { createContext, useContext } from 'react';
 
-export const ButtonGroupContext = createContext({});
+/** @type {import('./types').ButtonGroupContext} */
+const defaultContext = {};
+
+export const ButtonGroupContext = createContext(defaultContext);
```
saramarcondes, line 6: This is one way of typing a context, where you split the default context and cast it. The other was is below where you just type the return of createContext and TS will infer the type of the passed in default context.

packages/components/src/ControlGroup/ControlGroup.Context.js
```diff
@@ -1,4 +1,5 @@
 import { createContext, useContext } from 'react';
 
+/** @type {import('react').Context<import('./types').ControlGroupContext>} */
```
saramarcondes, line 3: Here's the example where the type of the default context is inferred.

packages/components/src/ControlGroup/ControlGroup.js
```diff
@@ -40,4 +44,26 @@ function ControlGroup(props, forwardedRef) {
...
+ *       <TextInput placeholder="First name" />
+ *       <Button variant="primary" />
+ *     </ControlGroup>
+ *   );
+ * }
+ * ```
+ *
+ * @type {import('@wp-g2/create-styles').PolymorphicComponent<'div', import('./types').Props>}
+ */
+const ConnectedControlGroup = contextConnect(ControlGroup, 'ControlGroup');
```
saramarcondes, line 67: We put the final description and example for the component on the connected one as that's what's actually exported, that way you get it in intellisense. If we put it on the actual component it would get lost.

packages/components/src/ControlGroup/useControlGroup.js
```diff
@@ -21,35 +24,46 @@ export function useControlGroup(props) {
...
-
-		return (
-			<ControlGroupContext.Provider key={_key} value={contextProps}>
-				{child}
-			</ControlGroupContext.Provider>
-		);
-	});
+	const clonedChildren =
+		validChildren &&
+		validChildren.map((child, index) => {
```
saramarcondes, line 29: This is one big whitespace change (disable whitespace changes to see) and just adds a conditional before mapping validChildren.

packages/components/src/Flex/useFlex.js
```diff
@@ -9,12 +9,12 @@ import * as styles from './Flex.styles';
  * @typedef FlexProps
  * @property {import('react').CSSProperties['alignItems']} [align='center']
  * @property {import('react').CSSProperties['alignItems']} [alignItems]
- * @property {import('../utils/types').ResponsiveCSSValue<import('react').CSSProperties['flexDirection']>} direction='row'
+ * @property {import('../utils/types').ResponsiveCSSValue<import('react').CSSProperties['flexDirection']>} [direction='row']
  * @property {boolean} [expanded]
- * @property {number} [gap]
+ * @property {number | string} [gap]
```
saramarcondes, line 14: Sometimes we pass in strings like `` '-1px' `` to `gap`.

packages/components/src/Flex/useFlex.js
```diff
@@ -9,12 +9,12 @@ import * as styles from './Flex.styles';
  * @typedef FlexProps
  * @property {import('react').CSSProperties['alignItems']} [align='center']
  * @property {import('react').CSSProperties['alignItems']} [alignItems]
- * @property {import('../utils/types').ResponsiveCSSValue<import('react').CSSProperties['flexDirection']>} direction='row'
+ * @property {import('../utils/types').ResponsiveCSSValue<import('react').CSSProperties['flexDirection']>} [direction='row']
```
saramarcondes, line 12: This and `wrap` were meant to be optional, mistakes that they weren't already.

packages/utils/src/react.js
```diff
@@ -6,8 +6,8 @@ export { default as mergeRefs } from 'react-merge-refs';
 export { default as hoistNonReactStatics } from 'hoist-non-react-statics';
 
 /**
- * @param {Parameters<React.ReactChildren['toArray']>} children
- * @return {ReturnType<React.ReactChildren['toArray']>}
+ * @param {import('react').ReactNode} children
+ * @return {import('react').ReactNodeArray}
```
saramarcondes, line 10: The original types for this were very lazy, these new types actually correspond to how we use this function in the codebase (always taking in `children` and always casting it to an array)

packages/components/src/Button/Button.js
ItsJonQ: TIL!

packages/components/src/ButtonGroup/ButtonGroup.Context.js
ItsJonQ: I see! I'm cool with this 👍

packages/components/src/ControlGroup/ControlGroup.js
ItsJonQ: Gotcha! I'm so thankful we have these initial components as example 🙏

packages/components/src/Flex/useFlex.js
ItsJonQ: True!

packages/utils/src/react.js
ItsJonQ: > The original types for this were very lazy
lolol. Thank you for the updates ❤️

ItsJonQ (approved):
🚀 from me! Thank you @saramarcondes !!!

PR #173, opened 2020-11-26
Type Card and its dependencies

saramarcondes:


ItsJonQ (approved):
🚀 from me!! These TS updates seem like they're getting easier (which is a good thing!) 😊

PR #174, opened 2020-11-26
Add typings to remake

saramarcondes:


ItsJonQ (approved):
🚀 !!! This is great :D.

Although... I don't think `remake` will work well moving forward, since we'll be using `kebab-case` for file names. 

Oh well. Maybe something to improve on in the future :)

PR #175, opened 2020-11-27
Type Dropdown and its dependencies

saramarcondes:
Also removes some (now) redundant type declarations after having improved the contextConnect type.


packages/components/src/Menu/Menu.js
```diff
@@ -1,23 +1,37 @@
...
+ */
 function Menu(props, forwardedRef) {
-	const {
-		children,
-		className,
-		menu: menuProp,
-		...otherProps
-	} = useContextSystem(props, 'Menu');
-
-	const { menu } = useDropdownContext();
```
ItsJonQ, line 20: Thank you for fixing this!

ItsJonQ (approved):
🚀 from me! Thank you @saramarcondes ! The improved `contextConnect()` types are so nice 😍

Issue #184, opened 2020-12-09
Utils: Resolve dependencies


ItsJonQ:
From the `utils` package, `array-move` is causing E2E issues for Gutenberg:
https://github.com/WordPress/gutenberg/pull/27594/checks?check_run_id=1520249117

On top of that, there could be things that could be refactored in the `utils` package to better prepare it for Gutenberg.
See this branch:
https://github.com/WordPress/gutenberg/tree/try/add-ui-utils-package

ItsJonQ:
Also! Adjust usage of `import _ from 'lodash'` to help with dead-code elimination

https://github.com/ItsJonQ/g2/blob/4c40577ad3e956f6ce5b35f0ac669a9eac744e7d/packages/utils/src/is.js#L1

PR #186, opened 2020-12-09
Utils: Update dependencies + lodash usage

ItsJonQ:
This update refactors the `utils` package to improve the shared dependencies with WordPress + Gutenberg.

Resolves https://github.com/ItsJonQ/g2/issues/184


ItsJonQ:
Hmm! Looks like there's some typescript build issues with importing the individual type checking utils from lodash:

```jsx
import {
  isArray,
  isBoolean,
  ...
} from 'lodash'
```

Refactored to do this:

```js
import * as _ from 'lodash';

const {
	isArray,
	isBoolean,
        ....
} = _
```

I'm not sure it makes a difference (treeshake/dead-code elimination wise).
The TS builder seems to be happy though.

PR #199, opened 2020-12-14
Remove automatic ref forwarding in core elements

saramarcondes:
### Testing instructions

run `yarn start`, open Storybook and ensure nothing is broken, especially in components that use forwarded refs like `TextInput`. Ensure refs of core elements are able to be retrieved (for eample in `TextInput` where we retrieve a ref of a `View` which is just the `div` core element).

### DIFF
--- a/packages/create-styles/src/create-style-system/create-core-element.js
+++ b/packages/create-styles/src/create-style-system/create-core-element.js
@@ -82,7 +82,6 @@ export const createCoreElement = (tagName, options) => {
 			as,
 			children,
 			className: classNameProp,
-			forwardedRef,
 			...props
 		},
 		ref,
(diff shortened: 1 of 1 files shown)


saramarcondes:
I'm not sure why Vercel failed, the build passes locally just fine 🤔 I'm not able to see anything when I follow the details link either. Maybe someone else has access that I don't? @jorgefilipecosta I've seen you contribute to parts of this project, do you have access to Vercel by any chance? Q is away for the holidays so I don't want to bug him.

jorgefilipecosta:
> @jorgefilipecosta I've seen you contribute to parts of this project, do you have access to Vercel by any chance?

Hi @saramarcondes, Unfortunately, I don't have access to vercel :(

packages/create-styles/src/create-style-system/create-core-element.js
```diff
@@ -136,16 +135,11 @@ export const createCoreElement = (tagName, options) => {
 			}
 		}
 
-		// eslint-disable-next-line
-		const refs = React.useMemo(() => {
-			return mergeRefs([forwardedRef, ref]);
```
gziolo, line 141: I have no idea if it's necessary but another fix could be as simple as:
```js
return forwardedRef ? mergeRefs([forwardedRef, ref]) : ref;
```
Maybe it would resolve the issue with Vercel?

gziolo:
It looks like it helped 😃

PR #211, opened 2021-01-08
Remove usages of is.function, is.string, and is.array.

saramarcondes:
In preparation of the migration, we can go ahead and refactor the usages of these is functions. I only did these three to prevent the PR from being too much to review.


ItsJonQ (approved):
🚀 from me. Thanks @saramarcondes 🤗 

I'm guessing we're keeping `is.defined` and `is.empty` because those are needed and can't really be done simply with a `typeof ===` check?

As we're dropping `is.`, those can be refactored to `isDefined` and `isEmpty`, similar to what you've done in the Gutenberg PRs

PR #221, opened 2021-01-12
create-styles: Use WP/data for use-reduced-motion hook

saramarcondes:
For some reason the tests fail, I have no idea why. Running storybook and manual testing works fine, but something is complaining about hook rules :( 

To manually test, run storybook and toggle reduced motion via the configuration options. Note that the change is effective by testing various components that use animation (like ZStack or Accordion)


ItsJonQ (approved):
🚀 from me! It looks like we're making progress in consolidating the state management workflows :)

PR #226, opened 2021-01-13
Remove zustand from color picker

saramarcondes:


ItsJonQ:
@saramarcondes Ahh!! I think I cracked it! Ultimately, I think it came down to how the `color` was being passed down into the ColorPicker library.

There was something about the `tinycolor2` transformed `rgba` string that was (sometimes) 1 value off from the one outputted from the `RgbaStringColorPicker` from `react-color`.

Because of that inconsistency, they keep trying to update each other, therefore, causing the loop.

(Still don't know why there would be an inconsistency? Maybe something to do with color algorithms... I don't know)

To fix this, I opted to use the `RgbaColorPicker`, which takes in `{r, g, b, a}`.
This alternative format seemed to make it happy.

Check out my working branch:

https://github.com/ItsJonQ/g2/blob/refactor/color-zustand-removal-2/packages/components/src/ColorPicker/useColorPicker.js#L141

(Note: I also merged in https://github.com/ItsJonQ/g2/pull/232 into `color-zustand-removal-2`)

ItsJonQ:
@saramarcondes Haiii!  I created a PR for my attempts at a 2nd pass:

https://github.com/ItsJonQ/g2/pull/235

PR #230, opened 2021-01-14
Refactor out zustand from select dropdown

saramarcondes:


saramarcondes:
There's something wrong with this, however, as `setCommitValue` causes an infinite loop. I've been debugging this for two hours and have made zero headway so pushing this up so I can focus on something else for a little bit and come back to this later.

ItsJonQ:
🤗  I just gave it a shot... no dice :(.

My hunch is that there's something with this from zustand:

```js
	const { commitValue } = store(
		(state) => ({ commitValue: state.commitValue }),
		shallowCompare,
	);
```

The `store()` (basically a hook) would only (force) update if the returned value is different, which can be tested with the final argument (`shallowCompare`). Perhaps this equality check prevented things from going infinite loop.

I tried this in your refactored version by using a couple of `ref` for values. I wasn't able to get it to work.

Alternatively, we could adjust the props <-> reducer state <-> downshift flow if this setup isn't work correctly.

As long as we're able to achieve the same features, that is for the component to be able to be controlled (synced) with preview/reset.

P.S. Zustand made all of this easier 😭 (IMO at least)

saramarcondes:
I think it's a race condition. If you console.log `commitValue` and `value` where `setCommitValue` is called in the `useUpdateEffect` you'll notice that it gets passed the _previous_ value each time, which is then committed becuase it doesn't match the committed value. This repeats, with the previous committed value being passed in as `value` each time.

I've traced the flow of data through this component many times and haven't been able to pinpoint how this is happening. It _seems_ like it should work just fine.

ItsJonQ (approved):
🚀 from me! @saramarcondes you're the best <3

Issue #203, opened 2020-12-29
Possible Simplifciation of useSelectDropdown


jorgefilipecosta:
Currently, useSelectDropdown uses a substate to keep state, but it seems all the state is synchronized from other entities that also keep the same state. So I guess we may be able to remove the store.

The state required from the SelectDropdown component is:
- the currently selected value.
- If the dropdown is open or not.
- The current item to highlight when the mouse moves.

The current value comes from the props. I think we should not have it as a local state. Currently, we have it as local state in two forms value, and commitValue. Why do we need a commit value can't we call onChange right away?

Regarding the dropdown open or not, and the current item to highlight state, it seems downshift provides a useSelect hook for this state, and we are using it, but then we synchronize it to our store. Can't we rely on downshift state as the source of truth?

I think the synchronization of state may be related to the issue https://github.com/ItsJonQ/g2/pull/201 and https://github.com/ItsJonQ/g2/issues/202.

ItsJonQ:
Resolved in https://github.com/ItsJonQ/g2/issues/223

PR #232, opened 2021-01-17
Improve value/state flow for TextInput and UnitInput

ItsJonQ:
This update improves how state (`value` from props) is handled, transformed, and derived for controlled components, such as `TextInput`, `UnitInput`, `Select`, `Slider`, and `SearchInput`.

This was originally brought to light by @jorgefilipecosta in https://github.com/ItsJonQ/g2/issues/203.
After ideas were shared within the Gutenberg integration PR, @saramarcondes and I took a closer look.

We started refactoring out `zustand`, which revealed some sync. issues with props/state.
We were successful in some refactors (such as `SelectDropdown`), but unsuccessful in others (such as `ColorPicker`).

The culmination of this journey (from `zustand` sync to recent refactors) has highlighted a good path forward for how we can reliably handle controlled components in a simple and stable fashion.

---

#### `useControlledValue()`

We needed an elegant and seamless way to handle value/state sync and defaultValue fallbacks for our controlled components. The solution I came up was this:

```js
/**
 * Simplified and improved implementation of useControlledState.
 *
 * @param {object} props
 * @param {string|number} [props.defaultValue]
 * @param {string|number} [props.value]
 * @param {() => void} [props.onChange]
 */
export function useControlledValue({
	defaultValue,
	onChange,
	value: valueProp,
}) {
	const hasValue = !isNil(valueProp);
	const initialValue = hasValue ? valueProp : defaultValue;

	const [state, setState] = React.useState(initialValue);

	const value = hasValue ? valueProp : state;
	const setValue = hasValue && !isNil(onChange) ? onChange : setState;

	return [value, setValue];
}
```

It's the successor to `useControlledState`.
It's pretty simple! If `value` is defined, use that. Otherwise, fallback to `state`.
(Same for `onChange`).

---

#### `usePropRef()`

Another improvement is the addition of `usePropRef`. This simulates the "selector" benefits that zustand provided for callback functions. By creating a ref for a collection of props, this collection can be passed into `useCallback` based functions without breaking dependency equality.

For `TextInput`, this ensures that `increment` and `decrement` doesn't change per render cycle. This is useful as actions (and others) are used throughout the component.

```js
/**
 * Creates a reference for a prop. This is useful for preserving dependency
 * memoization for hooks like useCallback.
 *
 * @example
 * ```js
 * // Referencing a simple prop, used in a useCallback function.
 * const valueRef = usePropRef(value)
 *
 * const increment = React.useCallback(() => {
 *   const value = valueRef.current
 *   onChange(value + 1)
 * }, [onChange, valueRef])
 * ```
 *
 * ---
 *
 * Multiple props can be passed in using in `object`.
 *
 * @example
 * ```js
 * const propRefs = usePropRef({ value, step })
 *
 * const increment = React.useCallback(() => {
 *   const { value, step } = propRefs.current
 *   onChange(value + step)
 * }, [onChange, propRefs])
 * ```
 */
export function usePropRef(prop) {
	const propRef = useRef(prop);

	useEffect(() => {
		propRef.current = prop;
	}, [prop]);

	return propRef;
}
```

From that, I refactored the previous `zustand` setup actions, selectors, etc... to callback functions and derived values (suggested by @jorgefilipecosta ).

From my tests, it looks like it worked!

---

#### Demo

The easiest place to test would be here:
https://g2-git-try-refactor-text-input-prop-state-handling.itsjonq.vercel.app/iframe.html?id=components-unitinput--default&viewMode=story

---

#### Examples

I've created a couple of CodeSandbox examples for the 2 new hooks

*useControlledValue*
https://codesandbox.io/s/use-controlled-value-kdpcg?file=/src/App.js

*usePropRef*
https://codesandbox.io/s/use-prop-ref-examples-4eowp

---

The big ones, that being `TextInput` and `UnitInput`, look like they're working as expected.

I suspect with these changes, we'll be able to complete the [ColorPicker refactor](https://github.com/ItsJonQ/g2/pull/226).
[From my tests](https://github.com/ItsJonQ/g2/tree/refactor/color-zustand-removal-2), I suspect the infinite looping issue is being caused by incompatible render cycles between controlled components.

🙏


packages/components/src/UnitInput/useUnitInputState.js
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			if (!validate(nextValue)) return;
+
+			onChange(nextValue);
+		},
+		[onChange, propRefs, validate],
+	);
+
+	const handleOnSelectChange = React.useCallback(
```
saramarcondes, line 115: Just a nit, but for new code, can we import `useCallback` directly from `React` rather than calling it off the `React` object? It just makes it easier to refactor in the future when we need to import those things from `@wordpress/element`.

packages/utils/src/hooks/use-controlled-value.js
```diff
@@ -0,0 +1,26 @@
+import { isNil } from 'lodash';
+import { useState } from 'react';
+
+/**
+ * Simplified and improved implementation of useControlledState.
+ *
+ * @param {object} props
+ * @param {string|number} [props.defaultValue]
+ * @param {string|number} [props.value]
+ * @param {() => void} [props.onChange]
```
saramarcondes, line 10: This should use a type parameter as it'll work for any value...
```suggestion
 * @template T
 * @param {object} props
 * @param {T} [props.defaultValue]
 * @param {T} [props.value]
 * @param {(value: T) => void} [props.onChange]
 * @return {[T, (value: T) => void]}
```

packages/utils/src/hooks/use-prop-ref.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function usePropRef(prop) {
+	const propRef = useRef(prop);
+
+	useEffect(() => {
+		propRef.current = prop;
+	}, [prop]);
+
+	return propRef;
+}
```
saramarcondes, line 40: Same here, let's use a type parameter so the type doesn't get erased to `any`.

saramarcondes (approved):
LGTM, just a couple things to think about

saramarcondes:
Good work sorting out a generic solution for this.

packages/components/src/UnitInput/useUnitInputState.js
ItsJonQ: > Just a nit, but for new code, can we import useCallback directly from React rather than calling it off the React object?

Can do! Thanks for pointing it out. That was something I forgot to refactor after my initial prototyped solution

PR #235, opened 2021-01-19
Remove zustand from color picker (Attempt 2)

ItsJonQ:
This is a branch for @saramarcondes 's PR https://github.com/ItsJonQ/g2/pull/226

---

We noticed some state update issues for the initial refactor attempt.

The prop/state/update flow has been simplified and refactored to match the conventions from:
https://github.com/ItsJonQ/g2/pull/232

Ultimately, I think it came down to how the color was being passed down into the ColorPicker library.

There was something about the tinycolor2 transformed rgba string that was (sometimes) 1 value off from the one outputted from the RgbaStringColorPicker from react-color.

Because of that inconsistency, they keep trying to update each other, therefore, causing the loop.

Still don't know why there would be an inconsistency? Maybe something to do with color algorithms... I don't know
https://github.com/omgovich/react-colorful/blob/master/src/utils/convert.ts

To fix this, I opted to use the RgbaColorPicker, which takes in `{r, g, b, a}`.

---

I don't think this could have happened without @saramarcondes 's initial attempt, coupled with our 🍐 sessions.
I drew a lot of insight and ideas from that work which helped the recent TextInput refactors as well as this one.


packages/components/package.json
```diff
@@ -28,7 +28,7 @@
     "history": "^4.9.0",
     "lodash": "^4.17.19",
     "path-to-regexp": "^1.7.0",
-    "react-colorful": "4.1.1",
+    "react-colorful": "4.4.4",
```
ItsJonQ, line 31: This gives us better a11y interactions (keyboard).
I think better performance/reliability as well.

packages/components/src/ColorPicker/useColorPicker.js
```diff
@@ -1,37 +1,157 @@
...
-import * as styles from './ColorPicker.styles';
-import { useColorPickerState } from './useColorPickerState';
+export function getColor(color) {
+	return ui.color(color).toRgbString();
+}
+
+/**
+ * @param {State} state
+ * @return {string}
+ */
```
saramarcondes, line 13: This is incorrect

packages/components/src/ColorPicker/useColorPicker.js
```diff
@@ -1,37 +1,157 @@
...
-import { noop } from '@wp-g2/utils';
+import { ui } from '@wp-g2/styles';
+import { noop } from 'lodash';
+import { useCallback, useState } from 'react';
 
-import * as styles from './ColorPicker.styles';
-import { useColorPickerState } from './useColorPickerState';
+export function getColor(color) {
+	return ui.color(color).toRgbString();
+}
```
saramarcondes, line 8: This exported function should be typed:

```suggestion
/**
 * @param {import('tinycolor2').ColorInput} color
 * @return {string}
 */
export function getColor(color) {
	return ui.color(color).toRgbString();
}
```

packages/components/src/ColorPicker/useColorPicker.js
```diff
@@ -1,37 +1,157 @@
...
+			break;
+	}
+
+	return colorValue;
+};
 
 export function useColorPicker(props) {
 	const {
-		color,
+		// @todo: Add support to interpret initial alpha + color.
```
saramarcondes, line 36: Does this todo still need to happen or is it out of scope for this PR?

ItsJonQ: I think it's out of scope for this PR 👍

ItsJonQ:
@saramarcondes Thanks for the initial review! I'll do my best to type/comment things 🙏

ItsJonQ:
@saramarcondes I tried! Haha

https://github.com/ItsJonQ/g2/pull/235/commits/c88f4984523faf87953e2250ec9a25c456b531c2

saramarcondes (approved):
LGTM!

PR #237, opened 2021-01-21
Packages: Align reakit dependency with WordPress packages

gziolo:
While working on https://github.com/WordPress/gutenberg/pull/28280, I discovered that Gutenberg uses 2 different versions of Reakit library. This PR tries to resolve that issue. I also removed `reakit-warning` dependency in favor of `@wordpress/warning` that is nearly identical and authored by the same mastermind @diegohaz 😄 

Screen Shot 2021-01-21 at 13 28 08

I don't know why it was pinned to v1.1.0. Let's make sure we can use the latest version.


ItsJonQ:
Thank you @gziolo !!

PR #239, opened 2021-01-21
SelectDropdown: Fix some a11y voiceover issues

ItsJonQ:
Screen Shot 2021-01-21 at 9 10 40 AM

This update fixes the a11y voiceover handling for the `SelectDropdown` component, which powers components like the `FontSizePicker`.

This was discovered after the integration of the `FontSizePicker` into Gutenberg:
https://github.com/WordPress/gutenberg/pull/27594

Screen Shot 2021-01-21 at 8 56 23 AM

The solution was the adjust the render structure for the `SelectDropdown` to better accommodate `useSelect()` from `downshift` (the same library/setup we use on the [`custom-select-control`](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/custom-select-control).

Unfortunately, we do lose the ability to `Portal` the menu content with this adjustment. It's more important that voiceover works correctly though.

We can revisit `useSelect` + portal'ing in the future.
From my experience, this doesn't seem like a very straight forward thing to do.

Other adjustments would be for the `TextInput`.
It's been adjusted so that `isCommitOnBlurOrEnter` is defaulted to `false`, matching the interaction flow for standard inputs and inputs found in the `@wordpress/components` library.

We can revisit the `input[type="number"]` interactions for `FontSizePicker` in the future.

---

### Testing

The quickest way to test would be to visit this URL:
https://g2-git-fix-select-dropdown-a11y.itsjonq.vercel.app/iframe.html?id=components-fontsizecontrol--periodic-rerender&viewMode=story

Turn on Mac OS voiceover to ensure that the focus and selections are working as expected.

--

cc'ing @talldan

### DIFF
--- a/packages/components/src/Alert/__tests__/__snapshots__/Alert.test.js.snap
+++ b/packages/components/src/Alert/__tests__/__snapshots__/Alert.test.js.snap
@@ -216,6 +216,16 @@ exports[`props should render as dismissable 1`] = `
 }
 
 .emotion-6.emotion-6.emotion-6.emotion-6.emotion-6.emotion-6.emotion-6 {
+  box-sizing: border-box;
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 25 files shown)


saramarcondes (approved):
LGTM!

Issue #245, opened 2021-01-22
Create Styles - Update to rtlcss v3 for PostCSS v8


ItsJonQ:
https://github.com/MohammadYounes/rtlcss/releases/tag/3.0.0

PR #246, opened 2021-01-22
Create Styles: Update rtlcss to the latest v3 for PostCSS v8

ItsJonQ:
Screen Shot 2021-01-22 at 11 08 20 AM

The easiest way to test would be to go into the `rtl.js` plugin and adjust the `return` condition to force RTL processing:

```js
function stylisRTLPlugin(context, content) {
	if (context === STYLIS_PROPERTY_CONTEXT) {
		// pass four undefineds to let TS know which overload we're using
		return true
			? rtlcss.process(content, undefined, undefined, undefined)
			: undefined;
	}
	return undefined;
}
```

Resolves https://github.com/ItsJonQ/g2/issues/245

### DIFF
--- a/packages/create-styles/package.json
+++ b/packages/create-styles/package.json
@@ -24,7 +24,7 @@
     "emotion-theming": "^10.0.27",
     "lodash": "^4.17.19",
     "mitt": "^2.1.0",
-    "rtlcss": "^2.6.2",
+    "rtlcss": "^3.0.0",
     "styled-griddie": "^0.1.3"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


gziolo:
I wanted to check if there is a way to switch locale in Storybook instead. I see there is an addon that would be useful for testing:
https://github.com/goooseman/storybook-addon-i18n

I haven't tested but I guess you can merge it based on the screenshot shared 👍🏻

gziolo (approved):

ItsJonQ:
@gziolo Thank you!! I do want to figure out how to improve the testing of RTL. And if possible (gasp) make it dynamic.

Issue #251, opened 2021-02-01
Layout: Improve Flex/Grid handling of longer labels


ItsJonQ:
<img width="287" alt="106353642-4a1ab400-62ec-11eb-8fdf-be147b44a106" src="https://user-images.githubusercontent.com/2322354/106486281-79adf580-647f-11eb-8c84-e9fa6d3be166.png">

Originally reported (https://github.com/WordPress/gutenberg/issues/28612)

---

The `FontSizeControl` uses CSS Grid to render it's inner UI. This works to keep the elements tidy. However, it restricts the ability to gracefully handle the rendering of longer translation labels (see screenshot above)

With some experimentation, I was able to adjust it to look like this:

<img width="553" alt="Screen Shot 2021-02-01 at 11 15 51 AM" src="https://user-images.githubusercontent.com/2322354/106486442-a6faa380-647f-11eb-9825-9dd94f73a529.png">
(Screenshot of Google translated German labels)

Original English labels:
<img width="579" alt="Screen Shot 2021-02-01 at 11 21 47 AM" src="https://user-images.githubusercontent.com/2322354/106486480-b11ca200-647f-11eb-96f8-e1e09e7f1707.png">

---

### Flex `wrap`

This involved swapping `Grid` -> `HStack`, and enhancing how `Flex`/`HStack` handles wrapping.

Previously, `Flex` only provided margin gaps on one axis (either `x` for horizontal, or `y` for vertical).
With the updated `<Flex wrap />`, the `gap` value is now applied on both `x` and `y` to accommodate wrapped content.

This solution seems to work well. However, it involves quite a bit of manual CSS for the individual children of `HStack`.

PR #252, opened 2021-02-01
Flex, HStack, FontSizePicker: Improve responsive rendering for longer translation labels

ItsJonQ:
This update improves Flex-based layout rendering for potentially longer translation labels.

This is achieved by enhancing how `Flex` handles `wrap` rendering. Previously, `Flex` only provided margin gaps on one axis (either `x` for horizontal, or `y` for vertical). With the updated `<Flex wrap />`, the `gap` value is now applied on both `x` and `y` to accommodate wrapped content.

This solution seems to work well. However, it involves quite a bit of manual CSS for the individual children of `HStack`.

The `FontSizeControl` how uses `HStack` instead of `Grid` for inner UI rendering. 

Screen Shot 2021-02-01 at 11 15 51 AM
(Screenshot of Google translated German labels)

Screen Shot 2021-02-01 at 11 21 47 AM
(Original English labels)

---

### FontSizeControl

The changes made to this `FontSizeControl` need to be ported to Gutenberg.
We technically don't need to update `FontSizeControl` here. I mostly did it for reference to make the updating easier in Gutenberg.

---

Resolves https://github.com/ItsJonQ/g2/issues/251

### DIFF
--- a/packages/components/src/font-size-control/font-size-control-styles.js
+++ b/packages/components/src/font-size-control/font-size-control-styles.js
@@ -1,8 +1,25 @@
-import { css } from '@wp-g2/styles';
+import { css, ui } from '@wp-g2/styles';
 
 export const FontSizeControl = css`
 	appearance: none;
 	border: none;
 	margin: 0;
 	padding: 0;
+`;
+
+export const SelectDropdownWrapper = css`
+	width: calc(100% * 0.5 - (${ui.space(2)} * 2));
+`;
+
+export const NumberInputWrapper = css`
+	width: calc(100% * 0.25 - (${ui.space(2)} * 2));
+`;
+
+export const ResetButtonWrapper = css`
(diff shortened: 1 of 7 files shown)


packages/components/src/font-size-control/font-size-control-styles.js
```diff
@@ -1,8 +1,25 @@
...
 
 export const FontSizeControl = css`
 	appearance: none;
 	border: none;
 	margin: 0;
 	padding: 0;
 `;
+
+export const SelectDropdownWrapper = css`
+	width: calc(100% * 0.5 - (${ui.space(2)} * 2));
```
ItsJonQ, line 11: I really don't like that we have to do this.

I don't see a way around it, unless we develop some sort of intelligent Grid inspired `templateColumn` feature for `HStack`.

PR #253, opened 2021-02-01
Panel: Fix visible prop handling

ItsJonQ:
This update fixes the visible prop rendering for Panels OUTSIDE of Accordions.

There was a bug in the `useAccordion` hook that prevent the `visible` prop from being correctly evaluated without rendering within an Accordion Context.


packages/components/src/Accordion/useAccordion.js
```diff
@@ -216,10 +216,12 @@ export function useAccordion({ id, visible: visibleProp }) {
 		set,
 	} = useAccordionContext();
 
-	const visible = getIsVisible(id);
+	const isWithinContext = set !== noop;
```
ItsJonQ, line 219: Probably not the best way to check 🤔.

saramarcondes: Definitely not, it leaks a huge implementation detail from `useAccordionContext`. Can we just default `set` to `null` instead?

ItsJonQ: I have a better idea :)! Pushing it up soon

saramarcondes:
@ItsJonQ What's the best way to test this?

ItsJonQ:
@saramarcondes Ah! My apologies 🙏 .

Make sure these Panels (Accordions) still sync with each other (when open/closing)
https://g2-git-fix-panel-initial-visible.itsjonq.vercel.app/iframe.html?id=components-accordion--default&viewMode=story

Ensure this Panel is initially expanded (and can be closed/opened)
https://g2-git-fix-panel-initial-visible.itsjonq.vercel.app/iframe.html?id=components-panel--default&viewMode=story

saramarcondes:
@ItsJonQ So it appears that the accordions only sync when they're opening, but they do not close together. Is that the intended behavior?

ItsJonQ:
>  but they do not close together. Is that the intended behavior?

🤦 no, haha. They should be closing together as well.

I think a good order would be to merge your https://github.com/ItsJonQ/g2/pull/255 update, and continue with this PR.

ItsJonQ:
@saramarcondes Investigated! It looks like it's an existing issue. I've added a todo for a fix. We can do it in a follow up :)

saramarcondes (approved):
Sounds good to me, approving in that case 😄

PR #255, opened 2021-02-02
Remove need for memoization in accordion

saramarcondes:
By moving action functions into module level functions we remove the need to memoize them. Their signature expands but not to the point of being unweildy.

Just an interesting approach to this problem, may or may not be something we adopt.

https://twitter.com/dan_abramov/status/1125774170154065920

### DIFF
--- a/packages/components/src/Accordion/Accordion.Context.js
+++ b/packages/components/src/Accordion/Accordion.Context.js
@@ -18,9 +18,7 @@ export const AccordionContext = createContext({
 	allowMultiple: false,
 	current: [],
 	getIsVisible: noop,
-	add: noop,
-	set: noop,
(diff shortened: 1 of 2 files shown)


ItsJonQ (approved):
🚀 from me! This is an interesting pattern :). I can see the benefits from it.
I think the requirement of passing `dispatch` as an argument is okay, as long as it's contained within the parent component or hook.

How it's used here is 👍 

I think where it would be cumbersome would be if the hook consumer would need to "assemble" the callback methods together when they want to use it (i.e. manually passing in `dispatch` as an arg).

For example, for a Reakit like workflow where the consumer creates components from hooks, assembling the callback methods + dispatch would feel verbose.

PR #265, opened 2021-02-19
fix: Spacing parameter types

sarayourfriend:
Match the type of these parameters with the type of the `space` parameter


ItsJonQ (approved):
🚀 Thank you @sarayourfriend !

PR #271, opened 2021-02-24
Add automagical interpolation of StyleQueries

sarayourfriend:
We invented another thing!


ItsJonQ (approved):
🚀 from me!!! Thank you @sarayourfriend

PR #273, opened 2021-02-25
Try interpolating styled components

sarayourfriend:
Just a start, we'll see what we can make of this.


ItsJonQ:
@sarayourfriend Yay!!! This was exactly the setup I was imagining :). But instead of :

```
// css.js
			if (isPolymorphicComponent(arg)) {
				return `[data-interpolation-name="${arg.__interpolationName__}"]`;
			}
```

When I was imagining this, I thought the CSS selector would be a custom CSS className of some kind.
Having a data attribute is basically the same thing 🤠 

Taking this for a spin now! ❤️

ItsJonQ (approved):
🚀 from me!!! @sarayourfriend Thank you for working on this 😍 .
I think the only thing left to do is to update snapshot tests 🙈

PR #276, opened 2021-03-01
Move divider dropdown styles

sarayourfriend:
Doing this so we don't forget to add these styles back to the dropdown when we migrate dropdown.


ItsJonQ (approved):
🚀 from me!! Thanks @sarayourfriend  :D. Very cool to take advantage of this new interpolation feature.

PR #277, opened 2021-03-01
Styled Components: Fix interpolated className

ItsJonQ:
This update consolidates how the interpolated className is created, in an attempt to resolve https://github.com/WordPress/gutenberg/pull/29084#issuecomment-788372734

Changes:

* A new `getInterpolatedClassName` util is created and exported from `@wp-g2/create-styles`
* `getInterpolatedClassName` ensures the `@emotion/hash` function is only called if the argument is valid
* `@wp-g2/context` adds `@wp-g2/create-styles` as a dependency (was previously missing)
* All calls to `@emotion/hash` go through the new `getInterpolatedClassName` util


ItsJonQ:
Going to try an alpha `0.0.160` release to test in Gutenberg

packages/create-styles/src/create-style-system/utils.js
```diff
@@ -167,3 +168,14 @@ export function compileInterpolatedStyles(interpolatedStyles, props) {
...
+
+/**
+ * Creates an interpolated className. This is used when styled components are
+ * interpolated in the template literal.
+ *
+ * @param {string} displayName The displayName of the component.
+ * @returns {string} The interpolated className.
+ */
+export function getInterpolatedClassName(displayName) {
+	return typeof displayName === 'string' ? `ic-${hash(displayName)}` : '';
```
sarayourfriend, line 180: Should we raise a development-env warning that the displayName was undefined?

sarayourfriend (approved):

packages/create-styles/src/create-style-system/utils.js
ItsJonQ: Good idea. But maybe not here. Perhaps in `contextConnect()` or `useContext()`?

What you think :)

sarayourfriend: Yup, that makes sense to me!

ItsJonQ: I just pushed that update :). Lemme know if that works

packages/context/src/context-connect.js
```diff
@@ -35,6 +35,12 @@ export function contextConnect(Component, namespace, options = {}) {
 		? namespace[0]
 		: namespace || WrappedComponent.name;
 
+	if (process.env.NODE_ENV === 'development') {
+		if (typeof namespace === 'undefined') {
+			console.warn('contextConnect', 'Please provide a namespace.');
+		}
+	}
+
```
sarayourfriend, line 43: We can use `@wordpress/warning` for this 🙂

ItsJonQ: We could, but I'd prefer not adding another `@wordpress/` dependency for now 🙈 haha

PR #279, opened 2021-03-04
Fix `getHeadingFontSize` types

sarayourfriend:
Allows passing `size="3"` OR `size={3}`, following the `ReactText` convention.


ItsJonQ:
Interesting! Check out the snapshot diff:

```
    -   font-size: calc(1 * 13px);
    -   font-size: calc(1 * var(--wp-g2-font-size));
    +   font-size: calc((13 / 13) * 13px);
    +   font-size: calc((13 / 13) * var(--wp-g2-font-size));
```

Mathematically, I think it's still the same :).
It looks like we'd be relying more on the CSS `calc`, which is fine :D

sarayourfriend:
@ItsJonQ Yeah, they're mathematically the same 😁 I'll update those snapshots, whoops!

PR #281, opened 2021-03-05
Do not use ForwardRefRenderFuction

sarayourfriend:
This simplifies the type and removes duplication of the `children` type which is problematic for elements like `hr` and `img` which do NOT have children.

Eventually we'll remove `children` from `ViewOwnProps` but that's a project for another day and will require a big refactor in the already ported component's types to explicitly annotate `children`. We should probably do this sooner than later though (probably this should be the first thing we do once these libraries exist in Gutenberg).


ItsJonQ (approved):
🚀 from me!

PR #284, opened 2021-03-10
Try: Reduce to just 1 level of specificity

sarayourfriend:
Just trying this out to see how it will interact with Gutenberg and the adapters.


sarayourfriend:
Merging this so we can try it out in GB

PR #288, opened 2021-03-15
Add type="button" by default to Button component

kevin940726:
`<Button>` component should have `type="button"` by default, or else it will default to `type="submit"` when placing inside a `<form>` element. It's also the default value in gutenberg's `<Button>` component: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/button/index.js#L77

Alternatively, we can enforce explicit `type` attribute with tools like [eslint](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/button-has-type.md).

I ran into this problem when trying to place some components here inside a `<form>` element: https://github.com/WordPress/gutenberg/pull/29755#issuecomment-797426815. IMHO, this should be handled in the design system rather in the app level.

### DIFF
--- a/packages/components/src/Alert/__tests__/__snapshots__/Alert.test.js.snap
+++ b/packages/components/src/Alert/__tests__/__snapshots__/Alert.test.js.snap
@@ -605,6 +605,7 @@ exports[`props should render as dismissable 1`] = `
           data-g2-component="CloseButton"
           data-icon="true"
           title="Dismiss"
+          type="button"
         >
           <span
             class="components-flex-item wp-components-flex-item ic-1p3tlc8 ic-192yz5d emotion-7"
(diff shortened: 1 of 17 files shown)


packages/components/src/Alert/__tests__/__snapshots__/Alert.test.js.snap
```diff
@@ -605,6 +605,7 @@ exports[`props should render as dismissable 1`] = `
           data-g2-component="CloseButton"
           data-icon="true"
           title="Dismiss"
+          type="button"
```
kevin940726, line 608: That's a lot of unnecessary snapshot files! Maybe we should just snapshot this at most once for each component and do the rests with normal asserting?

Ref: https://kentcdodds.com/blog/effective-snapshot-testing

ItsJonQ: @kevin940726 Great observation! For context, a majority of these tests were added as a "better than nothing" solution during the initial rapid development phase for this project :).

@sarayourfriend and I have been refactoring these as [we've migrated it over to Gutenberg](https://github.com/WordPress/gutenberg/issues/28399).

Speaking of which, we'll need to make this update in Gutenberg 👍

PR #293, opened 2021-03-23
Global Styles Sidebar (Design/Code V2)

ItsJonQ:
A revisited attempt to constructing the Global Styles sidebar experience:
https://github.com/WordPress/gutenberg/issues/27473

Created this draft PR to have auto deployed previews 🎉 

---

cc'ing @mtias @pablohoneyhoney @sarayourfriend

### DIFF
--- a/packages/components/src/Navigator/NavigatorLink.js
+++ b/packages/components/src/Navigator/NavigatorLink.js
@@ -78,6 +78,7 @@ function NavigatorLink(props, forwardedRef) {
 		);
 	}
 
+	console.log(to);
 	return (
 		<NavLink
 			{...otherProps}
--- a/packages/components/src/Navigator/Router/Link.js
+++ b/packages/components/src/Navigator/Router/Link.js
@@ -87,7 +87,10 @@ const Link = forwardRef(
 						context.location,
 					);
 
-					const href = location ? history.createHref(location) : '';
+					let href = location ? history.createHref(location) : '';
+					if (to) {
+						href = undefined;
+					}
 					const props = {
 						...rest,
 						href,
--- a/packages/components/src/Navigator/Router/NavLink.js
+++ b/packages/components/src/Navigator/Router/NavLink.js
@@ -86,7 +86,7 @@ const NavLink = forwardRef(
 						props.innerRef = innerRef;
 					}
 
-					return <Link {...props} />;
+					return <Link {...props} to={to} />;
 				}}
 			</RouterContext.Consumer>
 		);
--- a/packages/components/src/__fixtures__/GlobalStylesSidebarV2/GlobalStylesSidebarV2.js
+++ b/packages/components/src/__fixtures__/GlobalStylesSidebarV2/GlobalStylesSidebarV2.js
@@ -1,0 +1,97 @@
+import { ContextSystemProvider } from '@wp-g2/context';
+import { ui } from '@wp-g2/styles';
+import React from 'react';
+
+import {
+	CardHeader,
+	Heading,
+	Navigator,
+	NavigatorScreen,
+	NavigatorScreens,
+	View,
+} from '../../index';
+import { ColorsScreen, GlobalStylesScreen } from './screens';
+import { AppProvider } from './state/AppState';
+
+const ANIMATION_SPEED = 0.1;
+
+export default {
+	title: 'Examples/WIP/GlobalStylesSidebarV2',
+};
+
+const GlobalStylesHeader = () => {
+	return (
+		<CardHeader>
+			<Heading size={5}>Global Styles</Heading>
+		</CardHeader>
+	);
+};
+
+const screens = [
+	{
+		component: GlobalStylesScreen,
+		path: 'GlobalStyles',
+		title: 'Global Styles',
+	},
+	{
+		component: ColorsScreen,
+		path: 'Colors',
+		title: 'Colors',
+	},
+];
+
+const Sidebar = ({ children }) => {
+	return (
+		<ContextSystemProvider
+			value={{
+				Grid: { gap: 2 },
+				Icon: { size: 16 },
+			}}
+		>
+			<View
+				css={`
+					width: 280px;
+					position: absolute;
+					height: 100vh;
+					top: 0;
+					right: 0;
+					border-left: 1px solid ${ui.get('colorDivider')};
+				`}
+			>
+				{children}
+			</View>
+		</ContextSystemProvider>
+	);
+};
+
+const Example = (props) => {
+	const initialPath = 'Colors';
+	return (
+		<AppProvider>
+			<Navigator initialPath={initialPath}>
+				<GlobalStylesHeader />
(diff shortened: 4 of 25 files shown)


ItsJonQ:
The prototype code is located here:
`packages/components/src/__fixtures__/GlobalStylesSidebarV2/`

### 🧭 Routes

The idea behind the project structure is to construct it like a typical router-driven React application.
This is powered by the `Navigator` component, which is based largely on `react-router`.

### 📱 Screens

Continuing with the router paradigm, individual "views" are labelled as "screens" - borrowing a concept from mobile development. These screens are individual components that are coordinated through the main `Route` / `Switch`.

Not prototyped... In the eventual implementation, these screens may have individual loading/fetching states. In case they need to retrieve data from the global WP data store.

### ♻️ Get/Set

When rendering values from the store (e.g. paragraph color), I think it's profoundly important that values can be elegantly retrieved and updated as they are bound to controls.

We need to figure out a code pattern to support this.

Otherwise, it will get exponentially messy as we attempt to verbosely deconstruct/stitch together getter/setter values.
Not unlike how `attributes` and `setAttributes` works now within blocks.

ItsJonQ:
Just added a URL sync feature so that changes in the Sidebar screens (routes) is bookmarkable/sharable.

ItsJonQ:
### Potential (Navigator) Router conflicts

At the moment, any instance of `Navigator` will share the same context - the "brains" to coordinate the routing/navigation mechanics of the component.

I suspect we may run into issues if there are multiple nested `Navigator` components together.
For context, `Navigator` is basically the framework for creating generic (route-powered) navigation. It doesn't aesthetically resemble a sidebar, a carousel, etc... It's the framework that can be used to create any of those things.

In the case of Global Styles, we may have a carousel-like component rendering in the Sidebar.
And maybe this carousel component uses `Navigator`.
In this scenario, we may have conflicts.

### Solution: `createNavigator`?

Looking at the world of React Native, `@react-navigation` has a really nice (but fundamental) solution for this.
They have a `createStackNavigator` function, which generates the (pre-bound) contexts and components:
https://reactnavigation.org/docs/hello-react-navigation#creating-a-stack-navigator

In other words, we may need a (factory?) function to `createNavigator()`

Something like this:

```jsx
const GlobalStylesNavigator = createNavigator()
const { Navigator, NavigatorScreens, NavigatorScreen, useNavigator, ...theRest } = GlobalStylesNavigator
```

With the above implementation, we can have multiple Navigator-powered components living in harmony.

ItsJonQ:
cc'ing @griffbrad (Who may be interested in progress 🎉 )

ItsJonQ:
Screencast of latest(ish) updates:
https://www.loom.com/share/2cda11bb95174420b20e82ca96cf0892

I also recently added the bones for the Typography section:
Screen Shot 2021-03-24 at 2 55 18 PM

ItsJonQ:
### Control <-> Options

Screen Capture on 2021-03-25 at 16-00-25 (1)

Part of this new design introduces a new interaction to the sidebar controls. This dropdown consolidates both a "reset" mechanic and show/hide into a single interaction.

If we are to adopt this pattern to other style attributes, we have to come up with a solid (code) design pattern for how to handle this. From my experiments, the most consistent ways I've found to do this resolves around a handful of code patterns (Note: There may be better ones out there).

1. Rely heavily on abstracted `get`/`set`/`toggle` callbacks to pass in values to the control
2. Render the control is a wrapper that can automatically handle `show`/`hide`.
3. Using a `null` value for "show, but not set", and `undefined` for "hide"

---

Assuming we have something like this as the `styles` shape for a typography element:
```
{
  fontSize: '13px'
}
```

We'll walk through how we may connect that value with the UI with the design concepts above.
Note: The follow code examples will be high level pseudo code.

### get, set

The UI component may look like this:
```jsx
<FormGroup label="Font size">
  <UnitInput value={fontSize} onChange={handleOnChangeFontSize} />
</FormGroup>
```

The challenge is getting (or creating) the `fontSize` value and `handleOnChangeFontSize` callback.

Instead of manually destructuring (nested) objects and manually creating callbacks **per style value**, maybe we can have something like this:

```jsx
const [fontSize, setFontSize] = useStyleValue('typography.elements[0]')

<FormGroup label="Font size">
  <UnitInput value={fontSize} onChange={setFontSize} />
</FormGroup>
```

Borrowing from [`react-use-gesture`](https://use-gesture.netlify.app/docs/hooks/), maybe it could be simplified to just this:

```jsx
const bind = useStyleValue('typography.elements[0]')

<FormGroup label="Font size">
  <UnitInput {...bind()} />
</FormGroup>
```

Note: This would only work with the most straight-forward use-cases.

### Render

With out `value` and `onChange` bound to the UI control, we now need to conditionally render the `<FormGroup />` based on whether or not the style attribute is either enabled or set.

Instead of wrapping each UI chunk with something like:

```jsx
{fontSize !== undefined && (
  <FormGroup>
    ...
  </FormGroup>
)}
```

Maybe we can do something like this:
```jsx
<RenderControl prop="fontSize">
  <FormGroup>
    ...
  </FormGroup>
</RenderControl>
```

And the `RenderControl` renders the content depending on whether the `prop` is enabled / set.

### Dropdown / Options

Screen Shot 2021-03-25 at 4 52 51 PM

The dropdown of display options synchronize with the controls presented below.
These items will always show. The only thing that visually changes is the ✅ that indicates whether the controls is available or not.

With the current implementation of `<DropdownMenuItem />`, we just need to pass in a `boolean` value for `isSelected`:

```jsx
<DropdownMenuItem isSelected={isSelected} />
```

If we're going to be using this as a common design pattern, we'll need to generalize it enough so that we can do something like this:

```jsx
<Dropdown>
  <DropdownTrigger
    icon={<FiMoreHorizontal />}
    isControl
    isSubtle
    size="small"
  />
  <DropdownMenu>
    {options.map((option) => (
      <DropdownMenuItem {...option}>
        {option.label}
      </DropdownMenuItem>
    ))}
  </DropdownMenu>
</Dropdown>
```

Or better yet, something like this:

```jsx
<DisplayOptions options={[...]} />
```

### toggle

This happens when clicking an item from the display options control.

If we assume:

* `null` = Show control, but currently has no value (or enabled)
* `undefined` = Hide control (or disabled)

We can basically toggle the value like:

```
disable = undefined
enable = defaultValue || null
```

For the `DropdownMenuItem`, we should be using some sort of abstracted function as the callback. Something like this:

```jsx
<DropdownMenuItem onClick={toggleAttribute('fontSize')} {...options}>
  Font size
</DropdownMenuItem>
```

### defaultValue

Establishing a `defaultValue` of some kind is **important**. This is **not the same as initial value** for a control.
The `defaultValue` (as I'm describing it) is the value that would be toggled from `disabled` to `enabled`.
For example, from `undefined` (disabled) to `13px` (defaultValue).

Ideally, style attributes would have some sort of `defaultValue`. If not, we can use `null`.

---

Hope this helps!!

PR #295, opened 2021-03-26
Add preliminary documentation to create-styles

sarayourfriend:
Missing from this PR:

- [ ] ThemeProvider
- [ ] CSS Custom Properties (Can we delete this, it was for IE11 support right?)
- [ ] Hooks


ItsJonQ:
@sarayourfriend Thank you so much!!! 🙏 
This is a wonderful start! We could merge this PR now. We could also wait, if you want to get the other checklist items in.

(And yes, I think we can delete CSS Custom Properties :D)

PR #296, opened 2021-03-29
Remove CSS variable automatic fallback

sarayourfriend:


ItsJonQ (approved):
🚀 from me! Thanks @sarayourfriend :D

sarayourfriend:
I don't think we can merge this quite yet. Not sure when exactly Gutenberg is dropping IE11 :(

PR #298, opened 2021-04-01
Add `type="button"` to components using `as="button"` implicitly

kevin940726:
Similar to #288, this PR adds `type="button"` to those components implicitly using `as="button"`. So that they won't be rendered as a submit button when placed inside a `<form>` element.


ItsJonQ (approved):
Thanks @kevin940726 !! 🚀

